PR #28, opened 2015-02-03
Update to new FromStr

maurer:
What it says on the tin. Reports the ParseError instead of being an Option.

This will not build against nightly until a new nightly is produced - it is an update to deal with git master rustc.


alexcrichton:
Thanks! This repo currently tracks nightly instead of master but I'll be sure to merge this once there's a new nightly.

PR #35, opened 2015-04-11
Implement UUID version 5

Kroisse:
Because a SHA-1 hash library should be required to implement the UUID version 5, I described it as an additional feature on `Cargo.toml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ script:
 after_success: |
   [ $TRAVIS_BRANCH = master ] &&
   [ $TRAVIS_PULL_REQUEST = false ] &&
+  cargo doc --features="v5" &&  # build a full documentation
   echo '<meta http-equiv=refresh content=0;url=uuid/index.html>' > target/doc/index.html &&
   pip install ghp-import --user $USER &&
   $HOME/.local/bin/ghp-import -n target/doc &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: rust
 sudo: false
 script:
-  - cargo build --verbose --features="v5"
-  - cargo test --verbose --features="v5"
-  - cargo doc --features="v5"
+  - cargo build --verbose --features="$FEATURE"
+  - cargo test --verbose --features="$FEATURE"
+  - cargo doc --features="$FEATURE"
 after_success: |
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,6 +12,13 @@ description = """
 A library to generate and parse UUIDs.
 """
 
+[features]
+v5 = ["sha1"]
+
 [dependencies]
 rustc-serialize = "0.3"
 rand = "0.3"
+
```
alexcrichton: You could actually just add one new line here (which may be more concise):

```toml
sha1 = { version = "0.1.0", optional = true }
```

.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: rust
 sudo: false
 script:
-  - cargo build --verbose
-  - cargo test --verbose
-  - cargo doc
+  - cargo build --verbose --features="v5"
+  - cargo test --verbose --features="v5"
```
alexcrichton: Could you keep the original `cargo test` command as well? (just to make sure the lib works without the feature enabled)

alexcrichton:
Awesome, thanks!

Cargo.toml
Kroisse: I couldn't notice that it's possible. Thanks to your advice :D

.travis.yml
```diff
@@ -14,6 +14,9 @@ after_success: |
 env:
   global:
     secure: OiQj6NSyu634pS+qWCAeaibuBdiRsDMz5k0bi8hhdom3W24L8I9KkdWlBm2K8Ef97Uzgg6Xsr7LSWX0CONxX2eU3QaB1v3mXNxModa7ftW/w9y7B8MThxS5nRTUmHZC17OvoQ3EoNyrktnTAkyIr0E0YxBruJYAi4EH+54guRN8=
+  matrix:
+    - FEATURE=""
+    - FEATURE="v5"
```
alexcrichton: Unfortunately I think that this will cause the doc uploads to race with each other to see who wins. Perhaps just a new `cargo test --features v5` could be added?

fabricedesre:
@alexcrichton Are you still interested in that PR ?

alexcrichton:
Closing in favor of https://github.com/rust-lang-nursery/uuid/pull/54

PR #37, opened 2015-05-12
Add validation for uuid versions in parse_str

pipermerriam:
# What was the problem

According to seciton [4.1.3](https://tools.ietf.org/html/rfc4122#section-4.1.3) of the only valid versions for a uuid are `1, 2, 3, 4, 5`.  This was not being validated.

# What was fixed

Added an additional validation step to ensure that the version is one of the allowed versions.

# How can you see it

Take a look at the minor test suite modifications, or try to validate a uuid that has a non-valid version in it's 13th character.

*Here is a cute animal picture for your troubles*

bobomonkey

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,6 +139,7 @@ pub enum ParseError {
     InvalidCharacter(char, usize),
     InvalidGroups(usize),
     InvalidGroupLength(usize, usize, usize),
+    InvalidVersion(char),
 }
 
 /// Converts a ParseError to a string
@@ -157,6 +158,9 @@ impl fmt::Display for ParseError {
             ParseError::InvalidGroupLength(group, found, expecting) =>
                 write!(f, "Malformed; length of group {} was {}, \
                            expecting {}", group, found, expecting),
+            ParseError::InvalidVersion(version) =>
+                write!(f, "Invalid version; expecting 1, 2, 3, 4, or 5, \
+                           found {}", version),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kballard:
Is this actually something we should be doing? From that same RFC:

> Apart from determining whether the timestamp portion of the UUID
>       is in the future and therefore not yet assignable, there is no
>       mechanism for determining whether a UUID is 'valid'.

More generally, the version determines how the UUID is generated, and whether the timestamp field is valid or is randomly-generated (e.g. version 4 is just random). So there's value in being able to ask what the version of the UUID is, if you have some reason to want to inspect the fields, but I don't think there's any value in attempting to use this version to "validate" the UUID in any fashion. Most clients of UUIDs are using them purely as random identifiers, they don't actually care what the version is, much less whether it's one of the 5 known versions.

kballard:
Also, FWIW, if you do feel strongly that we should be validating, then there's more fields that need to be validated than just the version. The variant field has 2 valid values and 6 reserved values. Similarly, IIRC there's at least one other nibble that's restricted to a subset of values (at least for version 4; I'm not about to read through the whole RFC to verify this but that's my recollection).

Also, [§4.1.7](https://tools.ietf.org/html/rfc4122#section-4.1.7) states:

> The nil UUID is special form of UUID that is specified to have all
> 128 bits set to zero.

So the RFC itself acknowledges a particular form of UUID that doesn't match any of the known versions (or variants) and yet is still "valid".

pipermerriam:
@kballard In hind-sight I'm not convinced that there is (*much*) value in validating the version.  I can stretch for some very minor benefits like having `get_version` not have to return an `Option` type, but I'm new enough to rust to not have any gauge on whether that matters.

And since I can't think of a use case where I'd want to know that a uuid's version is invalid... I'm good just closing this.

Discovered rust a few days ago and since I'd already implemented this validation in another (python) library of mine, I seemed like a good learning exercise to see what it looked like in rust.  If you wouldn't mind, can you take a look at my actual code and let me know if there are better ways to do any of the things I did?

kballard:
@pipermerriam It looks reasonable to me, although I'm not sure why you're bothering to return early from a NULL UUID instead of just checking for NULL and allowing it if the version comes up as `'0'` (i.e. letting the rest of the code parse it like normal).

PR #39, opened 2015-05-30
Modify Debug and Display implementations to use to_hyphenated_string()

devyn:
Considering that the hyphenated format of UUIDs is the most common one, as far as I know, I think it's better for the `Display` implementation to output that format instead.

I've also replaced the `Debug` implementation, because reading `Uuid { bytes : [32, 46, 108, ...] }` in a log file isn't so nice.

I'll add a test for the `Debug` implementation if necessary.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,7 +102,7 @@ pub enum UuidVariant {
 }
 
 /// A Universally Unique Identifier (UUID)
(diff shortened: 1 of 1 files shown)


alexcrichton:
Sounds like a good idea to me, thanks!

PR #40, opened 2015-07-30
Fix documentation of to_simple_string()

agrover:
It is 16 octets but each octet takes two hex characters to represent, so the function returns
32 characters.


alexcrichton:
Thanks!

Issue #38, opened 2015-05-12
UUID parsing shouldn't need any allocation


kballard:
The UUID parsing routine right now uses at least three allocations: The vector of groups, the vector of group lengths, and the concatenated string of all hex digits. It should be easy to rewrite this to avoid any allocation whatsoever.

alexcrichton:
Fixed by @tomprogrammer in #42

PR #42, opened 2015-09-01
Parse

tomprogrammer:
The code iterates over the string after checking it's length and
stripping an urn prefix. The error reporting changes slightly because
errors are reported while iterating through the string. Anyway only the
first error is reported therefore the exact kind of error shouldn't
matter.

Error reporting changes:
 - The urn prefix is not counted in a `InvalidLength` error, only the
   Uuid part.
 - Invalid characters are reported as they appear and not checked for in
   advance.
 - `InvalidGroupLength` is checked for at every group boundary and at
   the end of the string.
 - At the end of the string both `InvalidGroupLength` and
   `InvalidGroups` can occur simultaneously since the length and all
   groups before were checked. In that case `InvalidGroups` is reported.

Simple benchmarks:
```
Before:
tests::bench_parse                   ... bench: 6,837 ns/iter (+/- 243)
tests::bench_parse_invalid_character ... bench:    42 ns/iter (+/- 4)
tests::bench_parse_invalid_group_len ... bench:   309 ns/iter (+/- 11)
tests::bench_parse_invalid_groups    ... bench:   274 ns/iter (+/- 6)
tests::bench_parse_invalid_len       ... bench:     0 ns/iter (+/- 0)
tests::bench_valid_hyphenated        ... bench:   628 ns/iter (+/- 27)
tests::bench_valid_short             ... bench:   460 ns/iter (+/- 17)

After:
tests::bench_parse                   ... bench: 1,502 ns/iter (+/- 61)
tests::bench_parse_invalid_character ... bench:    72 ns/iter (+/- 4)
tests::bench_parse_invalid_group_len ... bench:    98 ns/iter (+/- 4)
tests::bench_parse_invalid_groups    ... bench:   117 ns/iter (+/- 7)
tests::bench_parse_invalid_len       ... bench:     0 ns/iter (+/- 0)
tests::bench_valid_hyphenated        ... bench:   104 ns/iter (+/- 11)
tests::bench_valid_short             ... bench:    87 ns/iter (+/- 3)

Changes:
tests::bench_parse                   ... bench: ~ -5,335 ns/iter (~ 4.5×)
tests::bench_parse_invalid_character ... bench: ~    +30 ns/iter (~ 0.6×)
tests::bench_parse_invalid_group_len ... bench: ~   -211 ns/iter (~ 3.1×)
tests::bench_parse_invalid_groups    ... bench: ~   -157 ns/iter (~ 2.3×)
tests::bench_parse_invalid_len       ... bench:       -- --      (~---×)
tests::bench_valid_hyphenated        ... bench: ~   -524 ns/iter (~ 6.0×)
tests::bench_valid_short             ... bench: ~   -373 ns/iter (~ 5.2×)
```

Parsing the values is much faster now overall. Only one benchmark is
slower now, nevertheless it's faster than a successful parse. Only
`InvalidGroups` and `InvalidGroupLength` could be slighty slower if the
error occurs at the end of the string.


tomprogrammer:
Fixes #38

src/lib.rs
```diff
@@ -596,30 +609,122 @@ mod tests {
...
+            let _ = Uuid::parse_str("00000000000000000000000000000000");
+            let _ = Uuid::parse_str("00000000-0000-0000-0000-000000000000");
+
+            // Test error reporting
+            let _ = Uuid::parse_str("67e5504410b1426f9247bb680e5fe0c");
+            let _ = Uuid::parse_str("67e550X410b1426f9247bb680e5fe0cd");
+            let _ = Uuid::parse_str("67e550-4105b1426f9247bb680e5fe0c");
+            let _ = Uuid::parse_str("F9168C5E-CEB2-4faa-B6BF1-02BF39FA1E4");
+        });
+    }
```
alexcrichton: I've found a good way to organize benchmarks for now is to put them in the `benches` directory for Cargo. That way they're automatically picked up and we don't need an extra feature to get them to work on stable. It just means that `cargo bench` will only reliably succeed on nightly.

alexcrichton:
Could you also update `.travis.yml` to only run benchmarks on nightly? You can see [an example on the `rand` crate](https://github.com/rust-lang/rand/blob/master/.travis.yml#L11-L15)

src/lib.rs
tomprogrammer: Yes, that is much better and "forwards-compatible". Done

tomprogrammer:
I just copied the relevant part of `.travis.yml`, seems pretty straightforward.

tomprogrammer:
Could you please close #38, I didn't include that in my commit. :)

alexcrichton:
Sure!

PR #43, opened 2015-09-08
Update dependency `serde`

tomprogrammer:
Only a simple version update. No breakage expected because the feature `serde` wasn't released yet.


alexcrichton:
Thanks!

PR #46, opened 2015-10-09
formatted code with rustfmt

chills42:
This should have no functional changes, it simply reformats the code with the [rustfmt tool](https://github.com/nrc/rustfmt).

This goes along with the some of the work going on in rust. http://www.ncameron.org/blog/rustfmt-ing-rust/

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,7 +78,7 @@ use std::error::Error;
 use std::fmt;
 use std::hash;
 use std::iter::repeat;
-use std::mem::{transmute,transmute_copy};
+use std::mem::{transmute, transmute_copy};
 use std::str::FromStr;
 
 use rand::Rng;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -766,8 +787,22 @@ mod tests {
...
+                     0xc1,
+                     0xc2,
+                     0xd1,
+                     0xd2,
+                     0xd3,
+                     0xd4,
+                     0xd5,
+                     0xd6,
+                     0xd7,
+                     0xd8);
```
alexcrichton: cc @nrc, I've noticed in a few locations that this rule of one-expr-per-line-if-any-overflows ends up causing a loss in readability in some situations

alexcrichton:
Thanks @chills42!

src/lib.rs
nrc: I recommend using `#[rustfmt_skip]` for this sort of thing. They tend to be rare and hard for rustfmt to detect in general, so it is unlikely we'll be able to fix it in the short term.

Issue #48, opened 2015-10-19
0.1.18 Runtime Incompatibility


jnicholls:
I suggest creating an explicit unit test for string formatting of the UUID's Debug and Display impl.  0.1.17 used to_simple_string() to format Display, while 0.1.18 uses to_hyphenated_string(), which is an incompatible update under a compatible semver bump (0.1.17 -> 0.1.18).  The version should have been 0.2.0 or Display should have been left as is for 0.1.18.

Thanks.

alexcrichton:
I think this has since been dealt with, so closing.

PR #49, opened 2015-10-23
test: ensure formatting compatibility

chills42:
This adds two tests to address #48.


alexcrichton:
Can this perhaps just add a test for `Display` for now? Stability doesn't really cover `Debug` as it's basically always subject to change.

chills42:
Makes sense, removed the debug test.

alexcrichton:
Thanks!

PR #50, opened 2015-10-25
feat(from_bytes): return Result instead of Option

Icelle:
I thought it would make more sense to have `from_bytes` return a `Result` instead of an `Option`.

Let me know what you think!

[breaking-change]


alexcrichton:
Hm so I agree that this is the more appropriate signature for now, but I'm hesitant to merge as it's a breaking change. This may want to hold off until when this library is stabilized more generally.

Icelle:
Sounds good, thanks!

PR #51, opened 2015-11-20
Replace use of unwrap to improve safety

chills42:
This replaces the use of `.unwrap()` with `.ok_or(_)` to ensure
that a result is returned instead of allowing a possible panic.

A possible problem with this is that it does add an item to the ParseError enum, which is a breaking change.


alexcrichton:
Gah sorry! Not sure how this fell off the radar...

Thanks for the PR, though! It looks like it's impossible for this to return an error, however, so this is just an assertion that the code as written is correct, so I think the `unwrap` should stay.

PR #53, opened 2016-02-27
Update dependency `serde` to version 0.7.0

tomprogrammer:
This update isn't semver-compatible since it breaks users of `uuid` who use `serde = ^0.6.0`. Therefore we should raise the minior instead of patch I think. What is the common consensus in the rust ecosystem?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,5 @@ A library to generate and parse UUIDs.
 
 [dependencies]
 rustc-serialize = "0.3"
-serde = { version = "^0.6.0", optional = true }
+serde = { version = "^0.7.0", optional = true }
 rand = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -534,7 +534,7 @@ impl Decodable for Uuid {
 #[cfg(feature = "serde")]
 impl Serialize for Uuid {
     fn serialize<S: Serializer>(&self, serializer: &mut S) -> Result<(), S::Error> {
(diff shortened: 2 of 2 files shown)


kballard:
I don't know if there's a consensus on this, but surely anyone who has `serde = ^0.6.0` would just resolve to the latest version of `uuid` that's compatible with that constraint?

alexcrichton:
Thanks!

tomprogrammer:
Thanks for merging!

@kballard I agree with you, that anyone can pin `uuid` to the lastest working version, but that means the crate at first won't build (`uuid=0.1` as proposed by the readme) until the dependency is fixed (`uuid=0.1.18`). In my opinion that is not the idea of semver.

kballard:
@tomprogrammer You misunderstand. The dependency `uuid = "0.1"` is shorthand for `uuid = "^0.1"`, which itself is shorthand for `uuid = ">= 0.1, < 0.2"`. Cargo is supposed to find a set of package versions that satisfy all constraints. This means that if `uuid` version `0.1.18` requires `serde="^0.6"` and version `0.1.19` requires `serde="^0.7"`, and some other package in the dependency tree requires `serde="^0.6"`, then in theory Cargo should be resolving the `uuid="0.1"` dependency to version `0.1.18` instead of `0.1.19`, since `0.1.19` would produce incompatible versions for `serde`. Now I say "in theory" because I haven't actually tested its ability to backdate packages to handle incompatible version requirements of dependencies like this, but I would hope it does this correctly.

PR #54, opened 2016-03-02
uuid v5 implementation

fabricedesre:
Rebased https://github.com/rust-lang-nursery/uuid/pull/35

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,6 @@ after_success: |
 env:
   global:
     secure: OiQj6NSyu634pS+qWCAeaibuBdiRsDMz5k0bi8hhdom3W24L8I9KkdWlBm2K8Ef97Uzgg6Xsr7LSWX0CONxX2eU3QaB1v3mXNxModa7ftW/w9y7B8MThxS5nRTUmHZC17OvoQ3EoNyrktnTAkyIr0E0YxBruJYAi4EH+54guRN8=
-  matrix:
-    - FEATURE=""
-    - FEATURE="v5"
 notifications:
   email:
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks @fabricedesre! I'm a little worried about picking up a dependency on the `sha1` crate, however. It looks like [the implementation](https://github.com/mitsuhiko/rust-sha1/blob/master/src/lib.rs) has a lot of [unnecessary allocations](https://github.com/mitsuhiko/rust-sha1/blob/master/src/lib.rs#L117-L120) and in general may not quite be up to par with the nursery quality standards. Is there another possible implementation of sha1 we could use? Perhaps the sha1 calculation could be offloaded elsewhere?

fabricedesre:
Rust-crypto (https://github.com/DaGenix/rust-crypto/) has a sha1 implementation that looks solid, but I don't know what the criteria are for the nursery.

alexcrichton:
In the past there's been some mixed messaging about what the future of rust-crypto is. There may just unfortunately not be a great and small sha1 binding to use :(

alexcrichton:
Ok, I've [sent an upstream PR](https://github.com/mitsuhiko/rust-sha1/pull/13) to help get the library into shape as a dependency here, and then I think we can merge this.

There's also a few other minor breaking changes in the next release of uuid, so I'll release 0.2 with a few other tweaks as well.

steveklabnik:
This PR itself is now non-mergable

alexcrichton:
Ok, upstream got merged, so I'll merge this manually (belatedly)

PR #55, opened 2016-03-24
add to_bytes, for consuming uuid

viperscape:
Adds a method to consume Uuid and return inner bytes; alternative to referencing as_bytes. Thought it might be useful

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -326,6 +326,11 @@ impl Uuid {
         &self.bytes
     }
 
+    /// Return an array of 16 octets containing the UUID data, consumes Uuid
+    pub fn to_bytes(self) -> UuidBytes {
+        self.bytes
+    }
+
     /// Returns the UUID as a string of 32 hexadecimal digits
     ///
     /// Example: `936DA01F9ABD4d9d80C702AF85C822A8`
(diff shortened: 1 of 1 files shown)


alexcrichton:
Perhaps the `as_bytes` method could be changed to return `&[u8; 16]`? That way you could directly load from that (as it's a `Copy` type) to get the underlying bytes.

viperscape:
I don't know if that's considered a breaking change. I wanted a way to store the uuid as bytes without copying the bytes themselves, so just move semantics, it seemed like a good idea to just consume the uuid in this case. I'm not sure if that fits the library because it's awfully close to as_bytes, it's okay if it doesn't but thought I'd PR anyways :)

alexcrichton:
Due to coercions it's likely to not be a breaking change, but yeah there are subtle cases it could perhaps cause problems. If this should specifically be a *consuming* conversion then we probably want to call it `into_bytes` rather than just `to_bytes`, but the `Uuid` type is `Copy` anyway so you can't really consume an instance of it (it's just always copied)

alexcrichton:
Ok, with https://github.com/rust-lang-nursery/uuid/pull/56 on the horizon as a breaking change I'm gonna go ahead and change this to return a `&[u8; 16]` in `as_bytes` as another breaking change. Thanks for the PR though!

PR #56, opened 2016-03-31
Turn rustc-serialize support off by default.

Ms2ger:


alexcrichton:
Thanks! Can you also modify the travis config to test this? I'll also do the version bump as a separate commit, so it should be ok to leave that out for now.

Ms2ger:
@alexcrichton back to you

alexcrichton:
Thanks!

PR #58, opened 2016-04-04
Implement `Ord` and `PartialOrd` for `Uuid`

eeeeeta:
Useful as it allows a UUID to be used as a key for a BTreeMap, which wasn't possible before.

This is accomplished by using the underlying byte array's builtin Ord/PartialOrd implementations.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,7 +154,7 @@ pub enum UuidVariant {
 }
 
 /// A Universally Unique Identifier (UUID)
-#[derive(Copy, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Uuid {
     /// The 128-bit number stored in 16 bytes
     bytes: UuidBytes,


alexcrichton:
Thanks! Could this actually use `#[derive]`?

eeeeeta:
@alexcrichton Probably - I was just unsure how deriving an Ord impl works. Does it just use the struct members' implementations? If so, I'd be glad to just use that.

(I'm assuming you can just forcepush the pull request branch? I'm quite inexperienced with GitHub pull requests)

alexcrichton:
Yeah a force push should do the trick! And yeah `#[derive]` will basically "do the right thing" for aggregate structures (delegating to the implementation for each field)

eeeeeta:
@alexcrichton done! thank you for your help.

alexcrichton:
Thanks!

jgillich:
@alexcrichton Could you please publish a new release with this change? :)

alexcrichton:
Certainly! 0.2.1 published

jgillich:
Awesome, thanks!

PR #59, opened 2016-04-21
added Uuid v1 with random node support

liuchong:
added Uuid v1 with random node support
TODO: support generate node with a real MAC.

In fact in some situations such as in docker, the MAC is virtual. so I think to use a random node seem to be ok then.

please refer to: https://docs.docker.com/v1.8/articles/networking/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ A library to generate and parse UUIDs.
 rustc-serialize = { version = "0.3", optional = true }
 serde = { version = "^0.7.0", optional = true }
 rand = { version = "0.3", optional = true }
-time = "0.1"
 
 [features]
 use_std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,9 @@ A library to generate and parse UUIDs.
 rustc-serialize = { version = "0.3", optional = true }
 serde = { version = "^0.7.0", optional = true }
 rand = { version = "0.3", optional = true }
+time = "0.1"
 
 [features]
 use_std = []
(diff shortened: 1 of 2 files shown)


alexcrichton:
Unfortunately the `unsafe` code here isn't thread-safe, and it seems like trying to guess the MAC of the machine we're one is somewhat of a lost cause? Perhaps this could take the timestamp and the MAC as inputs to the constructor?

liuchong:
The "node" and "clock_seq" as given arguments are already in the fn "get_v1", I changed it's name to "new_v1_of". then added another commit to init the static_node use std::sync::{ONCE_INIT, Once}, is it ok that I've added an "extern crate std"?

alexcrichton:
Ah sorry, I mean that this should continue to not have a dependency on either `std` or `time`. Instead the `Uuid::new_v1` could be the only function added here and it could take two arguments, one for the MAC address and one for the timestamp perhaps?

liuchong:
https://github.com/rust-lang-nursery/uuid/pull/59/commits/9108a020ed2db95dfd62e4e97bf62d2436744333
Do you mean like this? (also, what do you think of if the clock_seq need to be in the parameters list?)

seems some usless, and it's almost the same as "from_fields".

alexcrichton:
Yeah I was thinking something like that (although also not having optional parameters which were in turn randomly generated).

I don't think we're really poised to fully provide V1 UUID generation because we don't have a robust method of figuring out what the local MAC address is?

liuchong:
close for now

PR #61, opened 2016-04-29
Fix index out of bounds panic in parse_str

### LINKED ISSUE #60: Index out of bounds panic in UUID parser
Hello,

I managed to cause an index out of bounds panic in the `parse_str` function with the following input:

```Rust
assert_eq!(Uuid::parse_str("231231212212423424324323477343246663"), Err(InvalidLength(36)));
```

The problem here is that the `parser_str` function checks for the input length (32 or 36), but does not have any "hyphen verification" in the 36 case. Because of this, the 33rd digit will cause the panic: 

[index out of bounds: the len is 16 but the index is 16', src/lib.rs:489](https://github.com/rust-lang-nursery/uuid/blob/0.2.1/src/lib.rs#L489)
```Rust
let mut buffer = [0u8; 16];
//...
buffer[(digit / 2) as usize] = acc;
//      ^^^^^^^^^
//         16
```

I tried to implement a quick fix in two different ways, but both have drawbacks.
I can count the hyphens when the input length is 36, or I can just check for the `digit` index variable in every iteration of the parser loop.

Which would be more efficient / cleaner solution? What kind of `ParseError` should I return in this case?

MrAnno:
Fixes #60

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -722,6 +722,8 @@ mod tests {
                    Err(InvalidLength(33)));
         assert_eq!(Uuid::parse_str("67e5504410b1426%9247bb680e5fe0c8"),
                    Err(InvalidCharacter('%', 15)));
+        assert_eq!(Uuid::parse_str("231231212212423424324323477343246663"),
+                   Err(InvalidLength(36)));
 
         // Valid
         assert!(Uuid::parse_str("00000000000000000000000000000000").is_ok());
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

MrAnno:
When will it be available on crates.io? :) (I can use my fork if the next release will be delayed.)

alexcrichton:
Ah no worries, I went ahead and published

MrAnno:
Wow, it was really fast. Thank you!

PR #63, opened 2016-05-29
lib: Change `rand` to `v4`

kraai:
The `rand` feature, which was renamed to `v4`, was still mentioned in the documentation.  Update those references.


alexcrichton:
Thanks!

PR #64, opened 2016-06-24
Improve some docs

steveklabnik:
They're not perfect yet, but add some examples, and fix some consistency issues

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -122,10 +122,10 @@ mod serde;
 #[cfg(feature = "v4")]
 use rand::Rng;
 
-/// A 128-bit (16 byte) buffer containing the ID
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -246,28 +264,65 @@ impl Uuid {
         }
     }
 
-    /// Creates a new random UUID
+    /// Creates a random `Uuid`.
+    ///
+    /// This uses the `rand` crate's default task RNG as the source of random numbers.
+    /// If you'd like to use a custom generator, don't use this method: use the
+    /// [`rand::Rand trait`]'s `rand()` method instead.
```
alexcrichton: Perhaps this could be expanded a bit as well to indicate that it requires enabling the `v4` feature of this library?

PR #68, opened 2016-07-07
Add doc examples for `uuid::Uuid::as_bytes`.

frewsxcv:


src/lib.rs
```diff
@@ -371,6 +371,20 @@ impl Uuid {
     }
 
     /// Return an array of 16 octets containing the UUID data
```
GuillaumeGomez: I just saw: we say octet in english? Isn't it byte?

GuillaumeGomez:
Seems good for me.

alexcrichton:
Thanks!

PR #72, opened 2016-09-02
`cargo install uuid` should give me a uuid tool.

nixpulvis:
We could consider adding options for version or other things in the future, but for now this suites my needs at least.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,7 @@ sha1 = { version = "0.2", optional = true }
 use_std = []
 v4 = ["rand"]
 v5 = ["sha1"]
+
+[[bin]]
+name = "uuid"
+doc = false
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,0 +1,11 @@
+extern crate uuid;
+
+use uuid::{Uuid, UuidVersion};
+
+fn main() {
+    let i = Uuid::new(UuidVersion::Random);
+    match i {
+        Some(name) => println!("{}", name),
+        None => panic!("unable to generate name")
+    }
+}


alexcrichton:
Thanks for the PR! I'd actually be surprised to have `cargo install uuid`, I'm curious why you think it should do something?

I'd also be hesitant to add a tool that *just* prints out a new random uuid, as we may want to consider what possible future behavior it would implement as well?

nixpulvis:
Maybe shouldn't have titled this with "should" but personally needed a UUID executable.

I was thinking it might be nice to provide most crates with simple executables, since installing them is opt in. This was the minimum functionality I could see for this crate.

I don't think it should be thought of as a tool that just prints a UUID, ideally it should do the things someone might want to do with UUIDs from the command line. I'm just not sure what all those things are.

alexcrichton:
Perhaps this could be in the `examples/` folder? That way you could do `cargo install uuid --example foo` to get this

nixpulvis:
I'd be ok with that. I like the idea of crates being an easy place for these kinds of small executables, but if we want to keep the clutter out of the default installs that's a good compromise.

I am curious what other binary might ever be used by `cargo install uuid` though.

alexcrichton:
True yeah, but we tend to err on the conservative side of things just to be safe.

nixpulvis:
The choice of `v4` vs `v5` is exposed with the standard options for cargo `--features=v4`, but I'm thinking there should be a default. I'm not sure which would be the best default.

lucab:
For reference, on several linux distribution the `uuid` command is already provided by http://www.ossp.org/pkg/lib/uuid/. If you are considering a generic-named installable utility, it may make sense to be CLI-compatible with https://linux.die.net/man/1/uuid.

nixpulvis:
@lucab I didn't know about `uuid` on some systems, I can look into that more. At first glance however it looks like the `-d` decode flag might require more work than this crate currently provides support for.

nixpulvis:
Relevent x-post from HN: https://segment.com/blog/a-brief-history-of-the-uuid/

nixpulvis:
Ok, so is not being complete really a reason not to merge this? This clearly does serve a purpose and is even written simply enough that it's expandable when and if we ever want to make the tool more complete.

radix:
I would suggest creating a new crate if you want to build a command-line program to generate UUIDs. You could just put this code in a `uuid-cli` crate or something?

Most unixy systems already have a a `uuid` tool, which has more options to do lots of different things. For this reason I think including it in this crate might be problematic, since it could conflict with that other version when a user didn't even expect to have it installed if they have it as a dependency.

kinggoesgaming:
Agree with @radix. The binary should be provided as a separate crate, which will allow adding more features, as well as allowing depending on other crates without maintenance burden for lib team.

PR #78, opened 2017-01-28
Upgrade to serde 0.9

rushmorem:
Serde 0.9 has some breaking changes in its traits. This pull request is an effort to make this library compatible with `serde` again.

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -1,35 +1,40 @@
 extern crate serde;
 extern crate std;
 
+use self::std::fmt;
 use self::std::prelude::v1::*;
 use self::serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
 use Uuid;
 
 impl Serialize for Uuid {
-    fn serialize<S: Serializer>(&self, serializer: &mut S) -> Result<(), S::Error> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         serializer.serialize_str(&self.hyphenated().to_string())
     }
 }
 
 impl Deserialize for Uuid {
-    fn deserialize<D: Deserializer>(deserializer: &mut D) -> Result<Self, D::Error> {
+    fn deserialize<D: Deserializer>(deserializer: D) -> Result<Self, D::Error> {
         struct UuidVisitor;
 
         impl de::Visitor for UuidVisitor {
             type Value = Uuid;
 
-            fn visit_str<E: de::Error>(&mut self, value: &str) -> Result<Uuid, E> {
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                write!(formatter, "a UUID")
+            }
+
+            fn visit_str<E: de::Error>(self, value: &str) -> Result<Uuid, E> {
                 value.parse::<Uuid>().map_err(|e| E::custom(e.to_string()))
             }
 
-            fn visit_bytes<E: de::Error>(&mut self, value: &[u8]) -> Result<Uuid, E> {
+            fn visit_bytes<E: de::Error>(self, value: &[u8]) -> Result<Uuid, E> {
                 Uuid::from_bytes(value).map_err(|e| E::custom(e.to_string()))
             }
         }
 
         deserializer.deserialize(UuidVisitor)
-                    .or_else(|_| deserializer.deserialize_string(UuidVisitor))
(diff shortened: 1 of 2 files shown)


src/serde.rs
```diff
@@ -1,35 +1,40 @@
...
-            fn visit_bytes<E: de::Error>(&mut self, value: &[u8]) -> Result<Uuid, E> {
+            fn visit_bytes<E: de::Error>(self, value: &[u8]) -> Result<Uuid, E> {
                 Uuid::from_bytes(value).map_err(|e| E::custom(e.to_string()))
             }
         }
 
         deserializer.deserialize(UuidVisitor)
-                    .or_else(|_| deserializer.deserialize_string(UuidVisitor))
-                    .or_else(|_| deserializer.deserialize_bytes(UuidVisitor))
+                    //.or_else(|_| deserializer.deserialize_string(UuidVisitor))
```
rushmorem: This is where I'm stuck. `deserializer` is no longer borrowed, `deserialize()` now consumes it so it's no longer available for the other `or_else` calls. All the tests pass with these commented out like this but it just doesn't feel right.

alexcrichton:
I think what'll need to happen here is that you call a deserializer method passing a visitor that handles the two cases. If the visitor is given a string, it parses it, and if the visitor is given a list of bytes, it parses that instead.

@dtolnay out of curiosity, what method should be called on the `D: Deserializer` first? Here the idea is that the impl supports decoding a `Uuid` out of a list of bytes or a string, where the string is parsed for a UUID and the bytes are just the literal bytes of the UUID. Should this start out by calling `deserialize`? Or perhaps `deserialize_string` to say that strings are "preferred" but other data types are still preferred?

dtolnay:
Neither - you should call `deserialize_str` because the UuidVisitor does not require ownership of the string data.

A helpful rule is to match whatever the Serialize implementation is doing. In this case Serialize calls serialize_str. If Deserialize calls deserialize then Bincode and other non-self-describing formats will not work. If Deserialize calls deserialize_bytes then Bincode and other non-self-describing formats will misinterpret the string that they wrote out - they serialized a string and they try to deserialize it as a byte array (maybe in Bincode this works but not in general).

The old implementation was never correct and is a perfect example of what the API changes in 0.9 were designed to prevent. Consider the following broken JSON data:

```json
"abc\q"abc"
```

In the old implementation of `Uuid::deserialize` it would start parsing this as a string, get as far as the `\q` which is not a valid JSON escape sequence and error out, then Uuid would call back into the deserializer and it would restart parsing at the second quotation  mark. All of a sudden an incorrect Deserialize implementation causes a correctly written Deserializer to accept invalid input.

alexcrichton:
Oh thanks for the info @dtolnay! Glad that the 0.9 changes are already weeding out bugs :)

Would you recommend dropping for both deserializing bytes-and-strings? Should this type only support deserializing strings? Or would deserialization of a list of bytes be supported by implementing an appropriate method in `Visitor` which would get called even though we called `deserialize_str` initially?

dtolnay:
There is already a visitor method there :smiley: - see UuidVisitor::visit_bytes. In this situation I don't know of any deserializer that would actually call it so I would be inclined to remove the visit_bytes method and just keep visit_str.

src/serde.rs
```diff
@@ -1,35 +1,40 @@
...
-    fn deserialize<D: Deserializer>(deserializer: &mut D) -> Result<Self, D::Error> {
+    fn deserialize<D: Deserializer>(deserializer: D) -> Result<Self, D::Error> {
         struct UuidVisitor;
 
         impl de::Visitor for UuidVisitor {
             type Value = Uuid;
 
-            fn visit_str<E: de::Error>(&mut self, value: &str) -> Result<Uuid, E> {
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                write!(formatter, "a UUID")
```
dtolnay: Change this to `"a UUID string"` to make the type information more helpful. For example in JSON if you try to read an integer as  UUID it will give you:

> invalid type: integer \`101\`, expected a UUID string at line x column y

rushmorem:
I have made some adjustments as per @dtolnay's recommendations, if I understood him correctly. David, did I get this right? If so, @alexcrichton let me know if you are OK with these changes so I can wrap up this pull request.

alexcrichton:
Thanks @rushmorem and @dtolnay!

rushmorem:
My pleasure :)

PR #81, opened 2017-03-04
Fix an index out of bounds panic in parse_str

nygrenh:
With specific malformed inputs, like: `"F9168C5E-CEB2F4faaFB6BFF329BF39FA1E4"`, `parse_str` used to panic with:

```
"thread '<unnamed>' panicked at 'index out of bounds: the len is 16 but
the index is 16', src/lib.rs:658"
```

The problem was in the loop that processes input characters: The loop happily writes all valid characters that aren't group delimiters to a buffer. With some cleverly placed hyphens, it was possible to fool the loop into trying to write more bytes to the buffer than is possible.

This problem was found with cargo-fuzz.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -605,8 +605,11 @@ impl Uuid {
         let mut buffer = [0u8; 16];
 
         for (i_char, chr) in input.chars().enumerate() {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #82, opened 2017-03-09
Update version numbers in docs to 0.4

ThomWright:
I got a bit confused about which version I should install when looking at the docs, hopefully this will help the next person :smile:


alexcrichton:
Thanks!

PR #85, opened 2017-04-20
Update to serde 1.0

lholden:
Deserialize and Visitor now requires a lifetime... just a small update to add this.


alexcrichton:
Thanks!

PR #92, opened 2017-06-07
Added v1 uuid generation and basic test

rrichardson:
This PR adds the V1 Uuid functionality.


src/lib.rs
```diff
@@ -299,6 +314,56 @@ impl Uuid {
...
+    }
+
+    /// Creates a thread-safe, internally mutable context to help ensure uniqueness
+    ///
+    /// This is a context which can be shared across threads.  It maintains an internal
+    /// counter that is incremented at every request, the value ends up in the clock_seq
+    /// portion of the V1 uuid (the fourth group).  This will improve the probability
+    /// that the UUID is unique across the process.
+    #[cfg(feature = "v1")]
+    pub fn new_v1_context(seed : Option<u64>) -> UuidV1Context {
```
alexcrichton: This seems like it may be best as a `new` method on `UuidV1Context` perhaps?

src/lib.rs
```diff
@@ -299,6 +314,56 @@ impl Uuid {
...
+    ///
+    /// The function is not guaranteed to produce monotonically increasing values
+    /// however.  There is a slight possibility that two successive equal time values
+    /// could be supplied and the sequence counter wraps back over to 0. 
+    ///
+    /// If uniqueness and monotonicity is required, the user is responsibile for ensuring
+    /// that the time value always increases between calls. 
+    /// (including between restarts of the process and device)
+    #[cfg(feature = "v1")]
+    pub fn new_v1(context: &UuidV1Context, time: u64, time_frac: u32, node: &[u8]) -> Result<Uuid, ParseError> {
```
alexcrichton: Perhaps `time_frac` could be `nanos: u32`? That way we could avoid floating math and such and it'd be a little clearer that this is a seconds/nanos pair.

Additionally, I think there's length restrictions on `node`, right? Could those be documented?

And finally, is it possible for this method to return an error? I figured it'd always generate a valid `Uuid`. If it can return an error though, can you document under what circumstances an error is returned?

src/lib.rs
```diff
@@ -205,12 +212,20 @@ pub const NAMESPACE_X500: Uuid = Uuid {
             0x80, 0xb4, 0x00, 0xc0, 0x4f, 0xd4, 0x30, 0xc8]
 };
 
+/// The number of 100 ns ticks between 
+/// the UUID epoch 1582-10-15 00:00:00 and the Unix epoch 1970-01-01 00:00:00
+pub const UUID_TICKS_BETWEEN_EPOCHS : u64 = 0x01B21DD213814000;
```
alexcrichton: Could this stay private for now?

rrichardson:
Requested changes have been made. Let me know if you need anything else.

src/lib.rs
```diff
@@ -205,12 +212,39 @@ pub const NAMESPACE_X500: Uuid = Uuid {
...
+impl UuidV1Context {
+
+    /// Creates a thread-safe, internally mutable context to help ensure uniqueness
+    ///
+    /// This is a context which can be shared across threads.  It maintains an internal
+    /// counter that is incremented at every request, the value ends up in the clock_seq
+    /// portion of the V1 uuid (the fourth group).  This will improve the probability
+    /// that the UUID is unique across the process.
+    pub fn new(seed : Option<u64>) -> UuidV1Context {
+        let initval = seed.unwrap_or(thread_rng().gen());
```
alexcrichton: This seems slightly odd perhaps? I'd imagine that `seed` is required to be passed in here or it's always randomly generated.

Additionally, this is then immediately truncated on 32-bit platforms?

alexcrichton: Perhaps this could take a `u16` (that seems to be all the used bits below) and then also avoid the usage of `rand`?

rrichardson: A u16 would make sense, I think. With regards to initializing to a random value, that is the norm for V1 implementations that I've seen.  I just wanted to provide a means to have continuuity between initialization of processes.

alexcrichton: True yeah. I wonder, how about something like this?

```rust
impl UuidV1Context {
    pub fn new(seed: u16) -> UuidV1Context;
}

impl Rand for UuidV1Context {
    // ...
}
```

That way they can be both explicitly requested? And we can document in `new` that you may wish to use `rand::random::<UuidV1Context>()` to generate a new context?

rrichardson: That is way awesome.

rrichardson: Oh wait..  Now I remember why I  I didn't go with u16 for the count:  Atomic u16 aren't in stable yet.  I have changed it to a usize (which should have the same wrapping behavior as a u16)  

If you want to go with a u16 and use the atomics, from nightly, I can do that as well.

rrichardson: Also I implemented Rand for UuidV1Context

alexcrichton:
@rrichardson oh I think it's ok to use `AtomicUsize` internally but I think the API should take a `u16`, right? An internal implementation detail could then just be to widen that up to a `usize`, right?

rrichardson:
Ok. I've changed it the public param to u16.
I think it's also be useful to have a method which unpacks the values back
out to a u64 (timestamp) and a u16 count. This would be for easy comparison.

It's only relevant for V1 uuids. What do you think?

alexcrichton:
Sure yeah sounds reasonable to me! We could have the methods return `Option` and only have it be `Some` for v1 uuids, right?

src/lib.rs
```diff
@@ -205,12 +210,54 @@ pub const NAMESPACE_X500: Uuid = Uuid {
...
+    /// Implementation for creating a context that has been initialized to a random value
+    ///
+    /// This is a context which can be shared across threads.  It maintains an internal
+    /// counter that is incremented at every request, the value ends up in the clock_seq
+    /// portion of the V1 uuid (the fourth group).  This will improve the probability
+    /// that the UUID is unique across the process.
+    fn rand<R: Rng>(rng: &mut R) -> Self {
+        UuidV1Context {
+            count: AtomicUsize::new(rng.gen())
+        }
```
nixpulvis: DRY?

rrichardson: about which part are you referring?  The documentation?

nixpulvis:
```rust
UuidV1Context::new(rng.gen())
```

nixpulvis: As for the docs... never blindly copy the docs of one function to another. This function's documentation can be very simple. Something like

> Builds (or returns) a random UuidV1 context, see `UuidV1Context::new` for more information.

rrichardson: I prefer explicit in the case of line 256,  but I agree,  the comments are repetitive.  Since I figured they'd appear in separate sections of the docs, I wanted to be clear about how to use it.  A reference, especially if I can convince the doc generator to make an anchor out of it, should suffice.

src/lib.rs
```diff
@@ -612,6 +716,32 @@ impl Uuid {
         Urn { inner: self }
     }
 
+
+    /// Returns an Optional Tuple of (u64, u16) representing the timestamp and 
+    /// counter portion of a V1 UUID.  If the supplied UUID is not V1, this 
+    /// will return None
+    #[cfg(feature = "v1")]
```
alexcrichton: I think we don't have to gate this behind the v1 feature, right? If you create a raw uuid from bytes you may still wish to parse it out even though you're not creating uuids?

alexcrichton:
Looks great to me, thanks so much again @rrichardson!

PR #93, opened 2017-06-12
Hash implementations from RustCrypto

newpavlov:
Crates from RustCrypto share the same interface and faster than `md5` (426 MB/s vs 336 MB/s) and `sha1` (490 MB/s vs 200 MB/s). And in future they could be even faster. (by [improving](https://github.com/rust-lang/rust/issues/41894) LLVM for MD5 and through hardware acceleration for SHA-1) For `uuid` it's probably not that important, but still improvement nevertheless.

The only drawback is that they depend on `typenum` (due to generic interfaces), but hopefully in future with the introduction of type-level integers this dependency will be removed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ all-features = true
 rustc-serialize = { version = "0.3", optional = true }
 serde = { version = "1.0", optional = true }
 rand = { version = "0.3", optional = true }
-sha-1 = { version = "0.4", optional = true }
-md-5 = { version = "0.5", optional = true }
+sha-1 = { version = "^0.4.1", optional = true, default-features = false }
+md-5 = { version = "^0.5.2", optional = true, default-features = false }
 
 [features]
 use_std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,7 @@ all-features = true
 rustc-serialize = { version = "0.3", optional = true }
 serde = { version = "1.0", optional = true }
 rand = { version = "0.3", optional = true }
-sha-1 = { version = "^0.4.1", optional = true, default-features = false }
-md-5 = { version = "^0.5.2", optional = true, default-features = false }
 
 [features]
 use_std = []
-v3 = ["md-5"]
 v4 = ["rand"]
-v5 = ["sha-1"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,11 @@ all-features = true
 rustc-serialize = { version = "0.3", optional = true }
 serde = { version = "1.0", optional = true }
 rand = { version = "0.3", optional = true }
-sha1 = { version = "0.2", optional = true }
-md5 = { version = "0.3", optional = true }
+sha-1 = { version = "0.4", optional = true }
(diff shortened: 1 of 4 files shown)


alexcrichton:
These seems to pull in some weighty dependencies, would it be possible to slim them down perhaps?

alexcrichton:
(also I'm not sure that speed is critical here?)

newpavlov:
I've feature-gated generic capabilities for md5 and sha1 (though this feature is enabled by default). This solution is not elegant, but it will do until we get type level integers.

BTW that is the policy on minimum supported Rust version for `uuid`?

alexcrichton:
Right now we support 2 stable versions back, but I still feel that these dependencies are a bit too weighty to pull in for the `uuid` crate :(

newpavlov:
Could you please elaborate why you feel this way? With disabled generic code those crates depend only on `byte-tools` (specialized version of byteorder), `block-buffer` and `arrayref`. All three crates are quite small. Of course, if you pull everything into one crate and will remove unused functions you will get minimal codebase, but I don't think that such approach is a good one.

 If I measured correctly (`cargo clean; cargo build --release`), compilation times are practically identical.

Well, if you'll continue to feel the same way, not much can be done about it, so in that case you can close this PR.

alexcrichton:
I feel this way because performance seems pretty non-critical here. The amount of data a UUID might hash seems super small? In that sense `uuid` is already pretty slim on dependencies and the intention is to keep it that way, so I'm not sure what the rationale for switching would be.

newpavlov:
Well, if you care about number of dependencies (not sure about reasons behind it considering cargo, but ok), how about pulling everything hash related into `uuid` by writing specialized functions? This will reduce number of dependencies even further from 5 to just 3.

alexcrichton:
Hm I'm sort of confused, why remove them? Aren't the dependencies just basically this code? I'm sort of getting lost as to the purpose of this PR...

newpavlov:
No, compress function for MD5 is taken from RustCrypto and better optimized, SHA-1 compress function is from `sha1`. (as more optimized version will take additional 300 lines of code) Also all unnecessary code is removed, so resulting code is minimal for `uuid` use case. (LOC and functionality wise)

I just followed your intent to keep number of dependencies minimal, if you think that 3 very small crates in dependency tree is too much, then removing 2 crates with additional optimization bonus should be a good thing, no?

As I said earlier, if you don't want change anything you can close this PR.

alexcrichton:
Ah ok, thanks for clarifying! I'm sort of skeptical about this though because it's vendoring an implementation for crates which already exist, and the implementation here is probably much less well tested than the external dependencies. If the implementation here is more efficient and/or better in some way, perhaps PRs to the upstream crates can be made?

newpavlov:
First: I've fully tested this implementation for data with sizes up to 10 KB using crates as a reference. Second: I've completely lost your position on matter of keeping number of dependencies small. Third: I've created PRs for both crates months ago, and both owners were against compatibility with RustCrypto hashes. (for whatever reasons)

To not waste no more of your and my time I will close this PR...

PR #94, opened 2017-06-12
Avoid doing character decoding when parsing uuids

Marwes:
Valid UUIDs are always ascii so we can avoid doing the decoding unless
we encounter a byte that is not a valid character.

```
 name                           orig ns/iter  new ns/iter  diff ns/iter   diff %  speedup
 bench_parse                    1,832         1,407                -425  -23.20%   x 1.30
 bench_parse_invalid_character  71            64                     -7   -9.86%   x 1.11
 bench_parse_invalid_group_len  102           80                    -22  -21.57%   x 1.27
 bench_parse_invalid_groups     123           100                   -23  -18.70%   x 1.23
 bench_parse_invalid_len        2             2                       0    0.00%   x 1.00
 bench_valid_hyphenated         116           96                    -20  -17.24%   x 1.21
 bench_valid_short              98            78                    -20  -20.41%   x 1.26

```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -631,7 +631,7 @@ impl Uuid {
         let mut acc = 0;
         let mut buffer = [0u8; 16];
 
-        for (i_char, chr) in input.chars().enumerate() {
+        for (i_char, chr) in input.bytes().enumerate() {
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍 

Nice wins!

PR #96, opened 2017-07-23
Eliminate a use of copy_memory

IvanUkhov:
It’s to avoid memory initialization before copying the digest.


alexcrichton:
Thanks!

PR #102, opened 2017-10-03
Add Uuid::as_fields() method

mdsteele:
This change adds a `Uuid::as_fields()` method, which is to `from_fields()` as `as_bytes()` is to `from_bytes()`.  (I found myself wishing for such a method when writing code elsewhere to serialize a `Uuid` with Windows byte order, i.e. each field separately little-endian.)  I tried to add new unit tests as closely analogous to those for `as_bytes` as possible (e.g. checking that `as_fields()` and `from_fields()` round-trip properly).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -649,19 +649,38 @@ impl Uuid {
 
     /// Returns the four field values of the UUID.
     ///
+    /// These values can be passed to the `from_fields()` method to get the
+    /// original `Uuid` back.
+    ///
+    /// * The first field value represents the first group of (eight) hex
+    ///   digits, taken as a big-endian `u32` value.  For V1 UUIDs, this field
+    ///   represents the low 32 bits of the timestamp.
+    /// * The second field value represents the second group of (four) hex
+    ///   digits, taken as a big-endian `u16` value.  For V1 UUIDs, this field
+    ///   represents the middle 16 bits of the timestamp.
+    /// * The third field value represents the third group of (four) hex
+    ///   digits, taken as a big-endian `u16` value.  The 4 most significant
+    ///   bits give the UUID version, and for V1 UUIDs, the last 12 bits
+    ///   represent the high 12 bits of the timestamp.
+    /// * The last field value represents the last two groups of four and
+    ///   twelve hex digits, taken in order.  The first 1-3 bits of this
+    ///   indicate the UUID variant, and for V1 UUIDs, the next 13-15 bits
+    ///   indicate the clock sequence and the last 48 bits indicate the node
+    ///   ID.
+    ///
     /// # Examples
     ///
     /// ```
     /// use uuid::Uuid;
     ///
     /// let uuid = Uuid::nil();
-    /// assert_eq!(uuid.as_fields(), (0, 0, 0, &[0u8; 8] as &[u8]));
+    /// assert_eq!(uuid.as_fields(), (0, 0, 0, &[0u8; 8]));
     ///
     /// let uuid = Uuid::parse_str("936DA01F-9ABD-4D9D-80C7-02AF85C822A8").unwrap();
     /// assert_eq!(uuid.as_fields(),
-    ///            (0x936DA01F, 0x9ABD, 0x4D9D, b"\x80\xC7\x02\xAF\x85\xC8\x22\xA8" as &[u8]));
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -647,6 +647,42 @@ impl Uuid {
...
+    pub fn as_fields(&self) -> (u32, u16, u16, [u8; 8]) {
+        let d1 = ((self.bytes[0] as u32) << 24) |
+                 ((self.bytes[1] as u32) << 16) |
+                 ((self.bytes[2] as u32) << 8) |
+                 (self.bytes[3] as u32);
+        let d2 = ((self.bytes[4] as u16) << 8) |
+                 (self.bytes[5] as u16);
+        let d3 = ((self.bytes[6] as u16) << 8) |
+                 (self.bytes[7] as u16);
+        let d4 = [
```
KodrAus: Is the motivation behind `[u8; 8]` here that it's more technically correct and smaller in size? Personally I think `&[u8]` is a better choice though given the `&self` signature and symmetry with `from_fields`. It would also simplify the implementation here too.

What do you think?

mdsteele: Mostly I was trying to match as closely as possible the signature of `as_bytes()`, which returns `&[u8; 16]`.  I would have liked to return `&[u8; 8]`, but I couldn't find a way to make that work, so I went with `[u8; 8]`.  But I'd be just as happy to go with `&[u8]`, if you prefer (and indeed it makes the implementation a bit simpler).

(The only other possible argument that occurs to me for returning `[u8; 8]` is that it's easier for the caller to turn that into `&[u8]` than it would be to turn a return value of `&[u8]` into a `[u8; 8]`, if the caller needs a static guarantee on the length for some reason.  But not sure how likely this is to matter -- for the use case I had in mind when I thought to add this method, a `&[u8]` works just fine.)

I'll update the PR.

src/lib.rs
```diff
@@ -647,6 +647,32 @@ impl Uuid {
...
 
+    /// Returns the four field values of the UUID.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use uuid::Uuid;
+    ///
+    /// let uuid = Uuid::nil();
+    /// assert_eq!(uuid.as_fields(), (0, 0, 0, &[0u8; 8] as &[u8]));
```
KodrAus: I don't think you need the `as &[u8]` here.

src/lib.rs
```diff
@@ -647,6 +647,32 @@ impl Uuid {
...
+    ///
+    /// let uuid = Uuid::parse_str("936DA01F-9ABD-4D9D-80C7-02AF85C822A8").unwrap();
+    /// assert_eq!(uuid.as_fields(),
+    ///            (0x936DA01F, 0x9ABD, 0x4D9D, b"\x80\xC7\x02\xAF\x85\xC8\x22\xA8" as &[u8]));
```
KodrAus: I don't think you need the `as &[u8]` here.

alexcrichton:
Nice! Could the documentation here briefly say what the fields are as well? (e.g. what each value of the tuple is). Also, I wonder if the slice return value may be best represented as `&[u8; 8]`?

KodrAus:
> Could the documentation here briefly say what the fields are as well?

This seems like a good idea :+1: We should do this for the [`from_fields`](https://docs.rs/uuid/0.5.1/uuid/struct.Uuid.html#method.from_fields) docs too I think.

> I wonder if the slice return value may be best represented as `&[u8; 8]`

We tried this originally, I don't suppose you can do this without some `unsafe` magic like:

```rust
let d4: &[u8; 8] = unsafe {
    let s = &self.bytes[8..16] as *const _ as *const u8;
    mem::transmute(s)
};
```

Is there a better way than that?

alexcrichton:
Nah yeah unfortunately there's no way to create a `&[u8; 8]` from a `&[u8]` safely but you could do something like:

```rust
fn promote(a: &[u8]) -> &[u8; 8] {
    assert_eq!(a.len(), 8);
    unsafe { &*(a as *const [u8] as *const [u8; 8]) }
}
```

alexcrichton:
er sorry, didn't mean to close

mdsteele:
I added some documentation for the fields (as best as I understand them) for `as_fields()`.

Thanks for the tips on using `unsafe` here; I've got it working now returning `&[u8; 8]`, which definitely seems nicer than before.

alexcrichton:
Thanks!

PR #103, opened 2017-10-07
Add impl fmt::{UpperHex,LowerHex} for {Uuid,Hyphenated,Simple}

sinkuu:
Not sure if this is useful, just for API completeness.


alexcrichton:
Thanks!

Issue #90, opened 2017-06-05
Is serde implementation going to be stable?


mulkieran:
Hi!

Are you going to stick with hyphenated format for all time?

radix:
There's no action item in this issue: @mulkieran is there something you're requesting be done? Does the merge of #104 resolve your concerns?

mulkieran:
Thanks for bringing this change to my attention. I think the short answer to my question _and_ yours is "No". But feel free to close the issue, I think we can deal.

PR #104, opened 2017-11-02
Serialize to binary if the serde format is not human readable

Marwes:
This uses the
[is_human_readable](https://github.com/serde-rs/serde/releases/tag/v1.0.16)
method added in serde 1.0.16 to serialize uuid's in a more compact
form if the serialized format does not need to be human readable.

This does not have an explicit test for `is_human_readable == false` since https://github.com/serde-rs/serde/issues/1065 has not been completed for serde_test so you may want to wait with merging. I have verified it manually though.

(Maybe?) related #90

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ md5 = { version = "0.3", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
-serde_test = "1.0.16"
 
 [features]
 use_std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,13 +17,14 @@ all-features = true
 
 [dependencies]
 rustc-serialize = { version = "0.3", optional = true }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Awesome, thanks! I wonder though, is this actually a breaking change? If the serialization changed for `IpAddr` then I'd imagine that this would be able to change as well, right?

Marwes:
Ah, right. It is only for *formats* setting `is_human_readable == false` that this is a breaking change (as reasoned in https://github.com/serde-rs/serde/pull/1044#issuecomment-327910843)

alexcrichton:
Ok! In that case I'm gonna merge this as it looks great!

Marwes:
Could we get a crates.io release with this?

(Is there a better way to request/indicate a crates.io release in stable/slow moving crates such as this?)

PR #105, opened 2017-11-07
Add tests for compact/readable representations

Marwes:
Ended up completing the compact/readable serde_test API so here are the
tests missing from https://github.com/rust-lang-nursery/uuid/pull/104.

I also replaced the `serde_json` roundtrip test with plain `assert_tokens`
tests which do essentially the same thing. I can re-add it if it seems
worthwhile to keep it though.


alexcrichton:
👍

PR #111, opened 2018-01-05
define rustc-serialize/serde as explicit features

ignatenkobrain:


ignatenkobrain:
Ugh....

```
  Features and dependencies cannot have the same name: `rustc-serialize`
```

Not sure what to do about this.. Having some features somewhere in code without declaring them seems really bad idea to me.

KodrAus:
Hi @ignatenkobrain! The error you're getting is because otpional crates are toggled using features, so this crate technically already has `rustc_serialize` and `serde` features because they're optional and redeclaring those features would be a conflict.

Is there a problem you're having that you're trying to solve here?

ignatenkobrain:
@KodrAus I package a lot of crates and we add Provides for each feature it provides sot that other crates do Requires on this features. Unfortunately, with rand there is no declared (in toml) serde feature so there is broken dependency..

ignatenkobrain:
(packaging in RPM I mean)

ignatenkobrain:
Opened https://github.com/rust-lang/cargo/issues/4911

KodrAus:
https://github.com/rust-lang/cargo/issues/4911 seems like the right place to keep talking about this since it's by-design behaviour in `cargo` we want to clarify more.

I'll go ahead and close this PR now, thanks again @ignatenkobrain!

PR #109, opened 2017-12-31
bump rand to 0.4

ignatenkobrain:

PR #112, opened 2018-01-24
Misc public api tidy ups

KodrAus:
After #109

* remove `rustc_serialize` support
* remove deprecated `use_std` feature
* make `rand` a private dependency (so we no longer implement `Rand` in the `v1` or `v4` features)
* fix some doc rendering issues


nox:
The removal of the `Rand` impl for `Uuid` removed the possibility to choose a different RNG, which is needed in Servo. Could the `rand` part of the changes be reverted, please? Cc @asajeffrey @eijebong

PR #119, opened 2018-01-31
Where's my boy IntelliJ

kinggoesgaming:
ignore rules for IntelliJ class of IDEs added


kinggoesgaming:
imma merge this as is, if any issues do arise, they can patched in the future

bors: r+

PR #130, opened 2018-02-03
use std by default

### LINKED ISSUE #123: Use `std` by default
Most users of `uuid` will be `std` users, anyone who is working on architectures that don't have support for `std` can disable the feature.

Steps needed:
* [x] Introduce the `default` key under `[features]` that has `["std"]` in Cargo.toml.
* [x] ~~Change CI environments so the change is reflected.~~ Not needed
* [x] Update documentation to reflect the changes.
* [x] Edit rust files to reflect the change.

kinggoesgaming:
closes #123

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,13 +26,10 @@
 //!
 //! # Dependencies
 //!
-//! By default, this crate depends on nothing but `libstd` and cannot generate
+//! By default, this crate depends on nothing but `std` and cannot generate
 //! [`Uuid`]s. You need to enable the following Cargo features to enable
 //! various pieces of functionality:
 //!
-//! * `std` - adds in functionality available when linking to the standard
-//!   library, currently this is only the `impl Error for ParseError`. This is
-//!   enabled by default.
 //! * `v1` - adds the `Uuid::new_v1` function and the ability to create a V1
 //!   using a `UUIDV1Context` and a timestamp from `time::timespec`
 //! * `v3` - adds the `Uuid::new_v3` function and the ability to create a V3
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,12 +26,13 @@
 //!
 //! # Dependencies
 //!
-//! This crate by default has no dependencies and is `#![no_std]` compatible.
-//! The following Cargo features, however, can be used to enable various pieces
-//! of functionality.
+//! By default, this crate depends on nothing but `libstd` and cannot generate
+//! any `Uuid`s. You need to enable the following Cargo features to enable
+//! various pieces of functionality:
 //!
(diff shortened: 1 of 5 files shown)


kinggoesgaming:
documentation still needs to be updated

src/lib.rs
```diff
@@ -26,12 +26,13 @@
 //!
 //! # Dependencies
 //!
-//! This crate by default has no dependencies and is `#![no_std]` compatible.
-//! The following Cargo features, however, can be used to enable various pieces
-//! of functionality.
+//! By default, this crate depends on nothing but `libstd` and cannot generate
```
KodrAus: nit: We could probably just say `std` instead of `libstd`.

src/lib.rs
```diff
@@ -26,12 +26,13 @@
...
+//! any `Uuid`s. You need to enable the following Cargo features to enable
+//! various pieces of functionality:
 //!
 //! * `std` - adds in functionality available when linking to the standard
```
KodrAus: We can probably just remove `std` from this list, since we're implying that you need to enable these features to get functionality,

src/lib.rs
```diff
@@ -103,7 +111,13 @@
...
 
 #![deny(warnings)]
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+// serde links to std, so go ahead an pull in our own std
+// support in those situations as well.
+#[cfg(any(feature = "std",
+          feature = "serde"))]
+extern crate core;
```
KodrAus: I don't think will work when compiling with `no_std`. Instead we might need to do something like:

```rust
#[cfg(feature = "std")]
extern crate std as core;
```

KodrAus: Actually, let's try and sort out our `std` features. Right now, if you specify `--no-default-features --features serde` you'll get a compile error from duplicate uses of `core`.

We never need anything from `serde` that uses `std` though, so we could specify it in our `Cargo.toml` as:

```toml
serde = { version = "1.0.16", optional = true, default-features = false }
```

And then include core in `lib.rs` as:

```rust
#[cfg(feature = "std")]
extern crate std as core;
```

Then `no_std` users can use `uuid` with `serde`.

kinggoesgaming:
@KodrAus this should ready to go unless you any more nitpicks :)

KodrAus:
Thanks @kinggoesgaming! Just one more comment:

We can change the attribute pulling in the `std_support` module to only require the `std` feature instead of also including `serde`.

kinggoesgaming:
@KodrAus done

KodrAus:
This looks good to me!

bors r+

Issue #106, opened 2017-11-17
V1 Generator Should not Always Increment Sequence Counter


posborne:
The V1 generation code (`new_v1`) rightly state the following in the documentation:

> This function is only guaranteed to produce unique values if the following conditions hold:
> 1. The NodeID is unique for this process.
> 2. The Context is shared across all threads which are generating V1 UUIDs
> 3. The supplied seconds+nsecs values are monotonically increasing.

Items 1 and 2 can be reasonably guaranteed in a system but the 3rd item is nearly impossible to guarantee in practice as the value needs to be monotonically increasing across the entire life of the system (not just a given boot).

Both the [RFC for UUID (RFC4122)](https://tools.ietf.org/html/rfc4122.html) and the [DCE documentation](http://pubs.opengroup.org/onlinepubs/9629399/apdxa.htm) provide similar guidelines on when the sequence should be modified.

I propose the following change (which I can create a PR for):

1. Make the `UuidV1Context` a trait so that the implementation can be plugged as needed.  This is important if one wanted to, for instance, ensure that changes to the time/sequence were persisted.
2. Make the `UuidV1Context` responsible for storing the most recent time value (in resolution of 100nsecs).
3. Add a method for getting an appropriate sequence number from a context given the current generation time.  As per the RFC recommendation, this sequence number would not change if the time was after the last time seen.  If the time was the same or prior to the last time seen, then the sequence number would be incremented.

Per RFC4122:

> 4.1.5.  Clock Sequence
> 
>    For UUID version 1, the clock sequence is used to help avoid
>    duplicates that could arise when the clock is set backwards in time
>    or if the node ID changes.
> 
>    If the clock is set backwards, or might have been set backwards
>    (e.g., while the system was powered off), and the UUID generator can
>    not be sure that no UUIDs were generated with timestamps larger than
>    the value to which the clock was set, then the clock sequence has to
>    be changed.  If the previous value of the clock sequence is known, it
>    can just be incremented; otherwise it should be set to a random or
>    high-quality pseudo-random value.

radix:
Hi, if you want to create a PR, I can help review it! (I have no authority in the uuid project, just offering my assistance)

radix:
Hi @posborne, do you still have any time to implement this change? If not, I can take a crack at it.

kinggoesgaming:
hi @posborne are you still interested in providing a PR for this? I want this to be part of 0.7 release. Else @radix can work on it.

radix:
[...]

PR #136, opened 2018-02-05
(wip for master branch freeze) WIP: Generalize sequence incrementing and only increment sequences when time doesn't move forwards

radix:
This is a big backwards-incompatible change, and WORK-IN-PROGRESS.

* `UuidV1Context` was renamed to `DefaultUuidV1Context`
* a new trait called `UuidV1Context` is added. This trait has a method for getting a u16 sequence number given the current time.
* `DefaultUuidV1Context` now has a Mutex in it for storing and updating the last time that a sequence number was requested. This means that `DefaultUuidV1Context` is only available when the `std` feature is enabled.

Is there any way we can avoid the context? I'm sure it probably slows down new_v1 a lot (though I plan on adding some benchmarks for this).

Also, while #106 proposed changing `UuidV1Context` to a trait, I think this might be unnecessary churn. A trait is a good idea, but it should probably be called something else so that `UuidV1Context` can remain backwards compatible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ cmake-build-debug/
 /target/
 Cargo.lock
 
+/.criterion
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ md5 = { version = "0.3", optional = true }
 
 [dev-dependencies]
 serde_test = "1.0.19"
+criterion = "0.1.0"
 
 [features]
 default = ["std"]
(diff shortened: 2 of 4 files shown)


kinggoesgaming:
> This means that DefaultUuidV1Context is only available when the std feature is enabled.

I think we need to think this one through a little more, there should be an option available when a user is in `no_std` mode.

radix:
@kinggoesgaming I have some other discussion points back on the ticket at #106.

kinggoesgaming:
superseded by #137

PR #142, opened 2018-02-08
Update sha1 to 0.5

### LINKED ISSUE #141: sha1 0.5
It would be very nice if you could update sha1 dependency to 0.5 before 0.6.

kinggoesgaming:
closes #141

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ all-features = true
 [dependencies]
 serde = { version = "1.0.16", optional = true, default-features = false }
 rand = { version = "0.4", optional = true }
-sha1 = { version = "0.4", optional = true }
+sha1 = { version = "0.5", optional = true }
 md5 = { version = "0.3", optional = true }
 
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
Kindly check if the repo has any breaking changes that affect us.

kinggoesgaming:
I will check the changes in between the two versions of rand

KodrAus:
I think we're safe either way here, because `sha1` isn't a public dependency.

kinggoesgaming:
The only major one I see is the switch to `JitterRNG` for `thread_local` so this should be good

Dylan-DPC:
bors: r+

ignatenkobrain:
I actually thought that you would reuse serde feature ;)

ignatenkobrain:
Also it has std, so you at least should change `std = []` → `std = [sha1/std]` or something like that

kinggoesgaming:
I can create a separate pr for that.. That was outside the scope of the issue

PR #145, opened 2018-02-11
Add from_uuid_bytes

ncallaway:
This adds a constructor that uses the explicit type `UuidBytes`.

The main difference between this constructor and `from_bytes` is that the length of `UuidBytes` is fixed at the correct length, so any use of `from_uuid_bytes` cannot fail. As such, instead of returning a `Result` it simply returns a `uuid`.

A secondary reason for this is that, if `const` functions ever move into nightly, this function would be an easy way to allow external consumers of this library to create `const` UUIDs.

If there are any other tests, or contribution guidelines that I've failed to include, please let me know.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -603,6 +603,41 @@ impl Uuid {
         Ok(uuid)
     }
 
+    /// Creates a `Uuid` using the supplied bytes.
+    ///
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// use uuid::Uuid;
+    /// use uuid::UuidBytes;
+    ///
+    /// let bytes:UuidBytes = [70, 235, 208, 238, 14, 109, 67, 201, 185, 13, 204, 195, 90, 145, 63, 62];
+    ///
+    /// let uuid = Uuid::from_uuid_bytes(&bytes);
+    /// let uuid = uuid.hyphenated().to_string();
+    ///
+    /// let expected_uuid = String::from("46ebd0ee-0e6d-43c9-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    /// An incorrect number of bytes:
+    ///
+    /// ```compile_fail
+    /// use uuid::Uuid;
+    /// use uuid::UuidBytes;
+    ///
+    /// let bytes:UuidBytes = [4, 54, 67, 12, 43, 2, 98, 76]; // doesn't compile
+    ///
+    /// let uuid = Uuid::from_uuid_bytes(&bytes);
+    /// ```
+    pub fn from_uuid_bytes(b: &UuidBytes) -> Uuid {
+        Uuid { bytes: b.clone() }
+    }
+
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -603,6 +603,41 @@ impl Uuid {
         Ok(uuid)
     }
 
+    /// Creates a `Uuid` using the supplied bytes.
+    ///
+    ///
```
kinggoesgaming: can you remove unnecessary whitespace? (Just to keep things consistent)

kinggoesgaming:
@ncallaway other than the review can you run `cargo +nightly fmt` aswell?

src/lib.rs
```diff
@@ -603,6 +603,40 @@ impl Uuid {
...
+    /// ```compile_fail
+    /// use uuid::Uuid;
+    /// use uuid::UuidBytes;
+    ///
+    /// let bytes:UuidBytes = [4, 54, 67, 12, 43, 2, 98, 76]; // doesn't compile
+    ///
+    /// let uuid = Uuid::from_uuid_bytes(&bytes);
+    /// ```
+    pub fn from_uuid_bytes(b: &UuidBytes) -> Uuid {
```
KodrAus: Since we just take a copy of these bytes, and `UuidBytes` is sized, maybe this could just take an owned `UuidBytes` instead of a `&UuidBytes` here? The clone will be really cheap, but it could be clearer that we take ownership of the raw bytes.

What do you think?

Dylan-DPC: `from_bytes()` also accepts a reference. So this keeps it consistent with that. Also I guess it is better to change the return type to a `Result` right?

KodrAus: Hmm, `from_bytes` takes a reference because it needs to, otherwise it's unsized. But `UuidBytes` isn't unsized, so we can take it by value if we want, since we need it by value.

I don't think we'd need to return a `Result`, because this always succeeds.

ncallaway: Yes, I deliberately didn't return a `Result`, because this always succeeds, as KodrAus noted. One motivation for this function (for me) is that I want to be able to create const Uuid's (if/when const functions come out of Rust nightly). I believe (thought I'm not positive) that if this returns a Result, this function wouldn't be able to become a const function that could generate a const Uuid.

I did have a reference in the signature to mirror `from_bytes()`, as noted by Dylan-DPC. However, I didn't have a strong motivational reason for this, other than matching the interface. I think KodrAus makes a good point that it might be better than the consumer of this function should be aware that a copy of the bytes are going to happen. As such, I'd be happy to change the interface to take an owned UuidBytes, and require the caller to perform the `.clone` if they want the copy.

I can make that change tomorrow morning (PST) and send the results. I can include the nightly rustfmt change at that time, too.

Dylan-DPC: Yeah i agree with it. Cool

ncallaway:
As discussed, this PR now has `::from_uuid_bytes` take an owned UuidBytes, instead of a reference. `cargo +nightly fmt` has been run as well.

Dylan-DPC:
bors: r+

PR #155, opened 2018-02-24
Defaults and enable features on rust playground

dns2utf8:
I tested this change with all features enabled and each one independently enabled.
Current state:

- [x] all enabled
- [x] independently
- [x] docs

I use this to test this PR:

```rust
extern crate uuid;

use uuid::Uuid;
use uuid::UuidVersion::*;

fn main() {
    let v = vec![ ("Mac", Mac), ("Dce", Dce), ("Md5", Md5), ("Random", Random), ("Sha1", Sha1) ];
    for variant in v {
        let my_uuid = Uuid::new(variant.1);
        println!("{:?} -> {:?}", variant.0, my_uuid);
    }
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,3 +36,8 @@ v1 = []
 v3 = ["md5", "rand"]
 v4 = ["rand"]
 v5 = ["sha1", "rand"]
+
+# Get features picked up by the Integer 32 playground. Not public API.
+#
+#    https://play.rust-lang.org
+playground = ["serde", "v1", "v3", "v4", "v5"]


KodrAus:
Thanks @dns2utf8! Do you have a case where you need a random v3 or v5 uuid? My understanding was that those types are typically used to get a consistent uuid value for some other identifier. The thing here I'm not sure of is defaulting to the `NAMESPACE_DNS` root, but maybe the default we'd use for random v3 and v5 uuids is a bit arbitrary either way.

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "uuid"
-version = "0.6.0"
+version = "0.6.1"
```
kinggoesgaming: We don't touch version number until preparing a release

dns2utf8: Done

dns2utf8:
hi @KodrAus 
When I read the docs I was under the impression the feature could be useful, so I don't have an immediate usecase apart from the learning effect on how features in rust work.

I stumbled on uuid after playing on the playground and I would like the playground to include all the features.

Dylan-DPC:
Hi. Thanks for the contribution but I still don't see the need for this. It just feels like we are adding more complexity for no gain.

dns2utf8:
One pro would be the crate will work as expected on https://play.rust-lang.org/?gist=118112c0ca3014d3683dfd6db7fa99ef&version=stable

KodrAus:
Ah I see! Yeh there's not a lot you can do without some of the version features enabled. Making `uuid` more useful on the playground sounds good to me, even without defaulting more uuid versions.

Is the `playground` feature just a conventional name that `play.integer32.com` picks up and includes?

shepmaster:
> Is the playground feature just a conventional name that play.integer32.com picks up and includes?

Yep, but both playgrounds have the same content.

KodrAus:
My thoughts are that we should go ahead with just the `playground` feature here, and then talk about values for other `Uuid::new` types in another PR.

What do you all think?

kinggoesgaming:
:+1: from me

Dylan-DPC:
Yeah @KodrAus I agree with that

Dylan-DPC:
@dns2utf8 can you update your branch with master?

kinggoesgaming:
@dns2utf8 can you merge master into defaults branch and fix the conglicts

kinggoesgaming:
@dns2utf8 I want this to be available in `uuid@0.6.3` can you merge changes from master and fix the conflicting changes?

PR #157, opened 2018-02-28
Bring back the renamed `use_std` feature for backwards compat

sgrif:
addaecd0fc9a1dbb053420ff91ed61ec1fc31395
removed this feature, which was intentionally left deprecated for
backwards compatibility. This "cleanup" removed a deprecated feature
before any released version actually had the deprecation present.

Removing this feature makes it impossible for other crates to support
UUID 0.6 without dropping support for UUID 0.5 and earlier. This is a
major problem for crates like Diesel, which are post-1.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ serde_test = "1.0.19"
 [features]
 default = ["std"]
 std = []
+use_std = ["std"] # Compatibility shim for 0.5 and earlier
 v1 = []
 v3 = ["md5"]
 v4 = ["rand"]


KodrAus:
Ah my bad! I thought that feature was already published in the deprecated state.

Do you mind adding a comment to the toml to note that it's a shim for compatibility between `0.4` and `0.5`?

sgrif:
Done!

kinggoesgaming:
Note though we will most likely remove it before 1.0. Should be enough time for crates to adapt

sgrif:
I would argue that keeping this compatibility shim is incredibly low maintenance, and is worth keeping given that this is a crate that is likely to be a public dependency of others. But ultimately it's your call :)

kinggoesgaming:
@sgrif I would love to have it removed in `1.0` maybe add note that this bound to be removed in the future.

Something like `use_std = ["std"] # Compatibility shim for 0.5 and earlier. May be removed in the feature. Use std instead`

Dylan-DPC:
hi @sgrif thanks for the PR. This is something that slipped through. We will maintain the compatibility shim in future releases.

> @sgrif I would love to have it removed in 1.0 maybe add a note that this bound to be removed in the future.

@kinggoesgaming yeah true but then users won't know about it. Not everyone will check the toml file. Instead we can put a warning in the docs & readme.

Idolf:
Wouldn't you be able to use the [semver-trick](https://github.com/dtolnay/semver-trick) to remove the feature flag from new releases while keeping backwards compatibility?

Issue #147, opened 2018-02-14
Use cfg_if for a cleaner sourceset


kinggoesgaming:
This should help in reading the sources a bit easier, as it helps in reading attributes in a more logical sense.

Reference: #124

kinggoesgaming:
forgot to tag it in PR #167

PR #167, opened 2018-03-11
Use cfg-if to create logical feature gate implementations

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [ ] feature enhancement
  - [ ] deprecation or removal
  - [x] refactor

# Description
`uuid` now adds `cfg-if` as a dependency.  With this, `#[cfg(..)]` implementations are easier to understand, as they can be handled in a more logical view because the macro allows us to `if` statements to make the checks more rustic at the source level

# Motivation
Understanding feature gate implementation becomes easier.

# Tests
The CI builds passing should be plenty.

# Related Issue(s)
#147

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -305,7 +305,7 @@ cfg_if! {
             }
         }
 
-        impl UuidV1ClockSequence {
+        impl UuidV1ClockSequence for UuidV1Context {
             fn generate_sequence(&self, _: u64, _: u32) -> u16 {
                 (self.count.fetch_add(1, atomic::Ordering::SeqCst) & 0xffff) as u16
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,10 +168,17 @@ cfg_if! {
     }
 }
 
-#[cfg(feature = "std")]
-mod std_support;
-#[cfg(feature = "serde")]
-mod serde_support;
+cfg_if! {
+    if #[cfg(feature = "serde")] {
+        mod serde_support;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


kinggoesgaming:
question: should I put various feature gated constructor `new` functions behind cfg_if` macros aswell or should we wait until a bigger refactor comes along?

kinggoesgaming:
@uuid-rs/uuid I think this is logical enough for this PR. Should be ready for review and merge

src/lib.rs
```diff
@@ -110,39 +110,75 @@
...
+cfg_if! {
+    if #[cfg(all(feature = "slog", not(test)))] {
+        extern crate slog;
+    } else if #[cfg(all(feature = "slog", test))] {
+        #[macro_use]
+        extern crate slog;
+    }
+}
+
+cfg_if! {
```
KodrAus: Maybe we could reduce some duplicated nesting here by conditionally using `core as std`:

```rust
if #[cfg(not(feature = "std"))] {
    use core as std;
}

use std::fmt;
use std::str;

...
```

What do you think?

kinggoesgaming: This PR was intended to switch to `cfg_if`. Another PR should make the that part more readable later... which I think I will have ready soon

Dylan-DPC:
bors: r+

kinggoesgaming:
clippy was outdated, cache cleared

bors r+

kinggoesgaming:
again! wrong cache cleared :(

bors r+

PR #168, opened 2018-03-15
make `rand` public again

### LINKED ISSUE #153: Custom RNGs
According to the [docs](https://docs.rs/uuid/0.6.0/uuid/struct.Uuid.html), it should be possible to use the `rand::Rand` trait's `rand()` method to generate a v4 UUID using a custom RNG:

> If you'd like to use a custom generator, don't use this method: use the rand::Rand trait's rand() method instead.

However, as of https://github.com/uuid-rs/uuid/commit/addaecd0fc9a1dbb053420ff91ed61ec1fc31395, `Uuid` no longer implements `rand::Rand` even if v4 support is enabled.  Is there some other mechanism to specify a custom RNG?

kinggoesgaming:
**I'm submitting a ...**
  - [x] bug fix
  - [ ] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
`rand` is once again declared as a public dependency.

# Motivation
Servo is currently uses custom `Rng` and as a result of #112, cannot use its own `Rng` implementation. Multiple other people also had this issue with the change in #112.

# Tests
N/A

# Related Issue(s)

#112
servo/servo#20306

closes #153

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,7 +121,7 @@ cfg_if! {
 
 cfg_if! {
     if #[cfg(feature = "rand")] {
-        extern crate rand;
+        pub extern crate rand;
     }
 }
 


KodrAus:
So the problem in https://github.com/servo/servo/issues/20306 isn't that the `rand` crate isn't publicly exported (which we definitely shouldn't do), it's that we removed an `impl Rand for Uuid`, which lets you pass in some `Rng` to generate `Uuid`s from.

kinggoesgaming:
I think @nox's suggestion here: https://github.com/servo/servo/issues/20306#issuecomment-373543459 is the best alternative to making `rand` public

KodrAus:
SO we should revert the `pub extern crate rand` change, and remove the bit of docs on `new_v4` about using the `rand::Rand` trait for a custom RNG.

kinggoesgaming:
I think I will close this PR over resolution that comes from #170

Dylan-DPC:
Closing as #170 will take care of this issue

Issue #170, opened 2018-03-17
Support V4 Uuids from arbitrary random bytes


KodrAus:
**Note: for support questions, please use [gitter] or [users forums]**. This 
repository's issues are reserved for feature requests and bug reports.

[gitter]: https://gitter.im/uuid-rs/Lobby
[users forums]: https://users.rust-lang.org

**I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request => Please do not submit support request here, see note 
  at the top of this template.

# Description

I thought we could flesh out our solution to creating uuids using alternative `Rng`s without having to make `rand` a public dependency.

## `new_v4_bat_scooter_name_not_important`

https://github.com/servo/servo/issues/20306 suggested a `new_v4_from_bytes` method as a workaround to needing to make `rand` a public dependency. Following the naming convention for other constructors, it might look like:

```rust
fn new_v4_from_bytes(b: &[u8]) -> Result<Uuid, ParseError>;
```

```rust
let uuid = Uuid::new_v4_from_bytes(b).expect("invalid length");
```

It would be nice to use `[u8; 16]` as the source instead of `&[u8]` so we can make the method infallible. It would be a bit inconsistent to use `fn new_v4_from_bytes(b: [u8; 16])` since we already have `fn from_bytes(b: &[u8])`. `new_v4_from_uuid_bytes` is a bit of a nasty name. So for now it could be:

```rust
fn new_v4_bat_scooter_name_not_important(b: [u8; 16]) -> Uuid;
```

```rust
let uuid = Uuid::new_v4_bat_scooter_name_not_important(b);
```

We could even go ahead and name it something like `new_v4_from_rng_bytes([u8; 16])` and deprecate it once `rand` is stable in favour of `Rand::rand`.

## `Builder`
Edit: Moved to a seperate issue #171 

Related to: #153 

cc: @nox @kinggoesgaming @Dylan-DPC

kinggoesgaming:
> We could even go ahead and name it something like `new_v4_from_rng_bytes([u8; 16])` and deprecate it once rand is stable in favour of `Rand::rand`.

I like this name, as it clearly gets the meaning across. One thing I think that should be added to this is that we add a `#[unstable] attribute so that users are aware why this function exists, as well as that it will be deprecated or removed when `rand` reaches `1.0`

nox:
I would use `random` instead of `rng` in the name. What do you mean by `#[unstable]`?

[...]

PR #172, opened 2018-03-17
from_rng_bytes added

Dylan-DPC:
**I'm submitting a ...**
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description

As discussed in the issue, this method is a solution to avoid making `rand` a public dependency

# Motivation
To avoid making `rand` public dependency

# Tests

none, test added 

# Related Issue(s)
#170


src/lib.rs
```diff
@@ -1804,6 +1831,20 @@ mod tests {
         assert_eq!(&b_in, b_out);
     }
 
+    #[test]
+    #[cfg(feature = "v4")]
+    fn test_from_rng_bytes(){
```
kinggoesgaming: Add space before `{`, rustfmt is complaining

src/lib.rs
```diff
@@ -685,6 +685,33 @@ impl Uuid {
...
+    ///
+    /// let expected_uuid = String::from("46ebd0ee-0e6d-43c9-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    #[cfg(feature = "v4")]
+    pub fn from_rng_bytes(b: [u8; 16]) -> Uuid {
+        let mut uuid = Uuid { bytes: [0; 16] };
+        copy_memory(&mut uuid.bytes, &b);
```
KodrAus: We also need to call `set_variant` and `set_version` here since the random bytes probably won't have the right values:

```rust
let mut uuid = Uuid { bytes: [0; 16] };
copy_memory(&mut uuid.bytes, &b);

uuid.set_variant(UuidVariant::RFC4122);
uuid.set_version(UuidVersion::Random);

uuid
```

KodrAus:
🚲🏚: I think I also agree with @nox's suggestion of using `from_random_bytes` instead of `from_rng_bytes`, since they're not really _random number generator bytes_, they're just _random bytes_.

nox:
Neither `from_rng_bytes` nor `from_random_bytes` convey that we are building a `v4` UUID, and those names are inconsistent with `Uuid::new_v4`.

KodrAus:
Hmm, I think `Uuid::from_random_bytes(b)` conveys that at least as well as `rng.gen::<Uuid>()` did.

src/lib.rs
```diff
@@ -685,6 +685,35 @@ impl Uuid {
...
+    ///
+    /// let uuid = Uuid::from_random_bytes(bytes);
+    /// let uuid = uuid.hyphenated().to_string();
+    ///
+    /// let expected_uuid = String::from("46ebd0ee-0e6d-43c9-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    #[cfg(feature = "v4")]
```
nox: There is no reason to feature-gate the method at all.

Dylan-DPC: We are only generating v4 uuids from this. So why have it on other features?

nox: Because there is no reason to have to bring in `rand` if one does not need it. For example in Servo we won't use `new_v4` at all (and thus the `rand` in `uuid`) because we need a better RNG.

Dylan-DPC: ok Cool.

src/lib.rs
```diff
@@ -685,6 +684,34 @@ impl Uuid {
...
+    /// let uuid = uuid.hyphenated().to_string();
+    ///
+    /// let expected_uuid = String::from("46ebd0ee-0e6d-43c9-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    pub fn from_random_bytes(b: [u8; 16]) -> Uuid {
+        let mut uuid = Uuid { bytes: [0; 16] };
+        copy_memory(&mut uuid.bytes, &b);
```
KodrAus: This wil need to be rebased now since `copy_memory` is now gone.

src/lib.rs
```diff
@@ -685,6 +684,34 @@ impl Uuid {
...
     }
 
+    /// Creates a v4 Uuid from rng bytes (ref issue #170)
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// use uuid::Uuid;
```
KodrAus: I still think we should add an example that's a little more motivating for people using this function, I don't think the value of the bytes we pass in, nor the assertion, should be visible in the remdered docs.

src/lib.rs
```diff
@@ -685,6 +684,34 @@ impl Uuid {
         Uuid { bytes: b }
     }
 
+    /// Creates a v4 Uuid from rng bytes (ref issue #170)
```
KodrAus: Probably don't need to reference GitHub issues in the docs prose.

src/lib.rs
Dylan-DPC: I just copied it from the examples given for the other functions. Will change it accordingly.

src/lib.rs
Dylan-DPC: sure. Since we have so many confusing functions to generate a UUID wanted to leave a note on why do we need this one :smile: will remove it

kinggoesgaming:
@KodrAus any further nits? Else can you approve?

src/lib.rs
```diff
@@ -685,6 +684,34 @@ impl Uuid {
...
+    /// let uuid = Uuid::from_random_bytes(bytes);
+    /// let uuid = uuid.hyphenated().to_string();
+    ///
+    /// let expected_uuid = String::from("46ebd0ee-0e6d-43c9-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    pub fn from_random_bytes(b: [u8; 16]) -> Uuid {
+        let mut uuid = Uuid { bytes: [0; 16] };
```
KodrAus: We don't need to create a new empty slice here and copy `b` into it, we could just create a `Uuid { bytes: b }`.

Dylan-DPC:
bors: r+

kinggoesgaming:
bors: r+

PR #176, opened 2018-03-19
Add Nil Version

### LINKED ISSUE #133: Add a Nil variant in UuidVersion enum
`0x00` version number is a valid value for UuidVersion, as per spec. This would make it required for users to explicitly check for the `nil` UUID.

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
Adds `Nil` variant to the `UuidVersion`.

# Motivation
`nil` `Uuid`, containing all zeros, is a valid `Uuid`, as per the specification. This will allow users doing pattern matching to explicitly check for it.

# Tests
Updated `test_nil` to ensure that 

# Related Issue(s)
closes #133

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -701,6 +701,7 @@ impl Uuid {
     pub fn get_version(&self) -> Option<UuidVersion> {
         let v = self.bytes[6] >> 4;
         match v {
+            0 => Some(UuidVersion::Nil),
             1 => Some(UuidVersion::Mac),
             2 => Some(UuidVersion::Dce),
             3 => Some(UuidVersion::Md5),
@@ -1279,6 +1280,8 @@ mod tests {
 
         assert!(nil.is_nil());
         assert!(!not_nil.is_nil());
+        assert_eq!(nil.get_version(), Some(UuidVersion::Nil));
+        assert_eq!(not_nil.get_version(), Some(UuidVersion::Random))
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -697,6 +701,7 @@ impl Uuid {
     pub fn get_version(&self) -> Option<UuidVersion> {
         let v = self.bytes[6] >> 4;
         match v {
+            0 => Some(UuidVersion::Nil),
```
KodrAus: If we're going to add an explicit `Nil` variant then I think we should change this match to `0 if self.is_nil()`, since this version is a bit unique in purely describing the value of the uuid. We can prove when the version is not valid. Otherwise:

```rust
let uuid = Uuid::from_uuid_bytes([4, 54, 67, 12, 43, 2, 2, 76, 32, 50, 87, 5, 1, 33, 43, 87]);

assert_eq!(nil.get_version(), Some(UuidVersion::Nil));
```

Is actually true, even though that uuid is definitely not nil.

kinggoesgaming:
@KodrAus I was actually thinking of removing the `Uuid::new` function. It serves no purpose other than for `v4` users.

src/lib.rs
kinggoesgaming: Fixed

kinggoesgaming:
RIP... wrong merge :( 

Opening a new PR

PR #181, opened 2018-03-21
Add Nil variant

### LINKED ISSUE #133: Add a Nil variant in UuidVersion enum

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
Adds `Nil` variant to the `UuidVersion`.

# Motivation
`nil` `Uuid`, containing all zeros, is a valid `Uuid`, as per the specification. This will allow users doing pattern matching to explicitly check for it.

# Tests
Updated `test_nil` to ensure that 

# Related Issue(s)
closes #133
supersedes #176

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,7 +151,9 @@ pub type UuidBytes = [u8; 16];
 /// The version of the UUID, denoting the generating algorithm.
 #[derive(Debug, PartialEq, Copy, Clone)]
 pub enum UuidVersion {
-    /// Special case for `nil` [`struct.Uuid.html`].
+    /// Special case for `nil` [`Uuid`].
+    ///
+    /// [`Uuid`]: struct.Uuid.html
     Nil,
     /// Version 1: MAC address
     Mac = 1,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -151,6 +151,8 @@ pub type UuidBytes = [u8; 16];
 /// The version of the UUID, denoting the generating algorithm.
 #[derive(Debug, PartialEq, Copy, Clone)]
 pub enum UuidVersion {
+    /// Special case for `nil` [`struct.Uuid.html`].
```
KodrAus: What do we have the `[`struct.Uuid.html`] here for? Does this render in the docs as `struct.Uuid.html`?

kinggoesgaming: Woops I forgot to link it will fix it ASAP

kinggoesgaming: That was supposed to be [`Uuid`](struct.Uuid.html)

kinggoesgaming: @KodrAus fixed.. can you approve

Dylan-DPC:
bors: r+

PR #189, opened 2018-03-30
Allow playground defaults

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
Allow compatibility with play.integer32.com and play.rust-lang.org

# Motivation
Allows people to use `uuid` in interactive code snippets

# Tests
Current tests passing

# Related Issue(s)
Supercedes #155


Dylan-DPC:
bors: r+

src/lib.rs
```diff
@@ -402,12 +402,30 @@ impl Uuid {
     /// Note that not all versions can be generated currently and `None` will be
     /// returned if the specified version cannot be generated.
     ///
+    /// To generate a random UUID (`UuidVersion::Md5`), then the `v3`
```
KodrAus: I think we should note in the docs that we're generating these random uuids off the well-known dns namespace uuid.

KodrAus: Hmm actually we probably don't need to, since we have enough randomness in the hash we jam into the uuid the base shouldn't matter so much.

PR #126, opened 2018-02-01
Added Appveyor build status button to the README file

gowthamsadasivam:
This fixes #122 
Added Appveyor build status button to the README file and linked the button to https://ci.appveyor.com/project/KodrAus/uuid

PR #198, opened 2018-04-19
Refactor/v4

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [ ] feature enhancement
  - [ ] deprecation or removal
  - [x] refactor

# Description
Move v4 impl into its own module

# Motivation
Part of the grand refactor effort

# Tests
N/A

# Related Issue(s)
#126


src/lib.rs
```diff
@@ -163,6 +163,18 @@ pub mod prelude;
...
 
+cfg_if! {
+    if #[cfg(feature = "v1")] {
+        pub mod v1;
+    }
+}
+
+cfg_if! {
+    if #[cfg(feature = "v4")] {
+        pub mod v4;
```
KodrAus: The `v4` module probably doesn't need to be public since we don't declare any public items in it, just an `impl` block. Otherwise it'll show up as an empty module in the docs.

Dylan-DPC:
bors: r+

PR #201, opened 2018-04-21
support u128

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
* features introduced: `u128`, `nightly`
* rust channel needed: `nightly`
* dependency introduced: `byteorder`

This introduced implementation for `u128` in the from of a `Uuid::from_u128` and `impl From`

# Motivation
`Uuid`s are 128 bits in size and rust natively provides a `u128` integer type. This 

# Tests
`u128_support::test_from_u128()` introduced and passes

# Related Issue(s)
N/A


Dylan-DPC:
bors: r+

src/u128_support.rs
```diff
@@ -0,0 +1,46 @@
+use byteorder;
+use prelude::*;
+
+impl Uuid {
+    /// Creates a new [`Uuid`] from a `u128` value.
+    ///
+    /// To create a [`Uuid`] from `u128`s, you need `u128` feature enabled for this crate.
+    ///
+    /// [`Uuid`]: ../struct.Uuid.html
+    #[inline(always)]
```
KodrAus: Usually `inline(always)` isn't ideal. A simple `inline` (without the `always`) would be better. Did you have a case where this method wasn't being inlined?

Issue #207, opened 2018-04-26
Add ability to configure default string format


ghost:
**I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

# Description
Add ability to configure default string format to use Simple instead of Hyphenated

# Feature Request

Currently, the default string format for Uuid is Hyphenated. Being able to override in a single place the default format to Single can reduce code duplication in projects which only use Simple.

PR #208, opened 2018-04-26
Add fmt_default_simple feature flag

codymack:
**I'm submitting a ...**
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
Possible implementation for #207 using conditional compilation feature flag.

# Motivation
For projects which consistently use Simple formatting and not Hyphenated formatting.

# Tests
Existing tests which test the default behavior of string formatting are modified to reference constants and variables which use the new feature flag.

# Other
I am new to Rust, so if anything in this PR is not good Rust practice, do let me know!

# Related Issue(s)

#207

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ v1 = []
 v3 = ["md5", "rand"]
 v4 = ["rand"]
 v5 = ["sha1", "rand"]
+fmt_default_simple = []
 
 # Get features picked up by the Integer 32 playground. Not public API.
 #
--- a/src/core_support.rs
+++ b/src/core_support.rs
@@ -174,8 +174,13 @@ mod tests {
 
         assert_eq!(s.len(), FMT_LENGTH);
 
-        check!(buffer, "{}", s, FMT_LENGTH, |c| c.is_lowercase() || c.is_digit(10)
-            || c == '-');
+        check!(
+            buffer,
+            "{}",
+            s,
+            FMT_LENGTH,
+            |c| c.is_lowercase() || c.is_digit(10) || c == '-'
+        );
     }
 
     #[test]
--- a/src/core_support.rs
+++ b/src/core_support.rs
@@ -27,15 +27,31 @@ impl fmt::Display for UuidVariant {
     }
 }
 
-impl fmt::LowerHex for Uuid {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        <super::Hyphenated as fmt::LowerHex>::fmt(&self.hyphenated(), f)
-    }
-}
+cfg_if! {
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -39,6 +39,7 @@ v1 = []
 v3 = ["md5", "rand"]
 v4 = ["rand"]
 v5 = ["sha1", "rand"]
+fmt_default_simple = []
```
kinggoesgaming: add a `fmt_default_hyphenated` and add it in the feature `default`

src/core_support.rs
```diff
@@ -27,15 +27,31 @@ impl fmt::Display for UuidVariant {
     }
 }
 
-impl fmt::LowerHex for Uuid {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        <super::Hyphenated as fmt::LowerHex>::fmt(&self.hyphenated(), f)
-    }
-}
+cfg_if! {
```
kinggoesgaming: what if two different libraries of a project are using different defaults?

src/test_util.rs
```diff
@@ -1,5 +1,13 @@
 use Uuid;
 
+cfg_if! {
```
kinggoesgaming: wait on one of my PRs that is adding a public accessible constants for this

kinggoesgaming: wait for #212

Cargo.toml
kinggoesgaming: actually `default_fmt_hyphenated` and `default_fmt_simple` would do better

KodrAus:
Hmm, I'm not sure feature flags are the best approach here because cargo features are additive, and they don't line up well with features that replace functionality. That means you can get surprising behaviour if different dependents of `uuid` in a single dependency tree use different formatting features.

@codymack do you have a case where you need a specific format for all displayed uuids?

src/test_util.rs
Dylan-DPC: @kinggoesgaming #212 is targetted to 0.7.0 and this is for 0.6.4

kinggoesgaming: Let's see if I can retarget the branch

kinggoesgaming:
In a PR coming soon I will be providing adapters that are serde compatible so this may not even be needed

codymack:
I am going to close this PR. After thinking about this more, feature flags (which, as you say, are additive) are not a good solution, due to possible conflicting dependencies.

KodrAus:
Thanks for the PR though @codymack! If you've got a case where you're finding yourself duplicating a lot of code to format uuids the way you need feel free to leave some more details on #207 and we can try hash out an alternative solution.

PR #218, opened 2018-05-04
Add const Uuid::from_uuid_bytes

mokeyish:
**I'm submitting a  const Uuid::from_uuid_bytes for creating const uuid value.**  [ref uuid_macro](https://github.com/uuid-rs/uuid_macro/pull/1)
  - [ ] bug fix
  - [x] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description

# Motivation

# Tests

# Related Issue(s)


src/lib.rs
```diff
@@ -111,6 +111,7 @@
 )]
 #![deny(warnings)]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![feature(const_fn)]
```
Dylan-DPC: this is not allowed for the stable build. you will have to wrap it in a `cfg_if`

src/lib.rs
```diff
@@ -111,7 +111,8 @@
 )]
 #![deny(warnings)]
 #![cfg_attr(not(feature = "std"), no_std)]
-#![feature(const_fn)]
+#![cfg_attr(feature = "nightly", feature(const_fn))]
```
kinggoesgaming: write this as `#![cfg_attr(nightly, feature(const_fn))]`

Cargo.toml
```diff
@@ -33,6 +33,7 @@ serde_test = "1.0.19"
 
 [features]
 default = ["std"]
+nightly = []
```
kinggoesgaming: I don't think we need this for this atm.. see the other comment below

mokeyish: @kinggoesgaming It is not working when I remove ```nightly = []``` and use ```#![cfg_attr(nightly, feature(const_fn))]```

kinggoesgaming: tested it locally.. its works for me... have you tried `cargo +nightly build`? it needs that `+nightly` bit... if you get something like toolchain not found error... run `rustup install nightly` and try again

Dylan-DPC: travis passes as well on nightly means it is working

mokeyish: @kinggoesgaming @Dylan-DPC Why is it only ```test_uuid_macro_build```test passed? The ```test_const_uuid_macro_build``` should also be passed.
image

kinggoesgaming: hmm its not even being detected/run, which is indeed weird... 

have you pushed these changes to your fork?

mokeyish: @kinggoesgaming It's didn't work as expectedly, So I haven't pushed it yet.

mokeyish: @kinggoesgaming I pushed it on [uuid_macro/nightly-uuid](https://github.com/uuid-rs/uuid_macro/tree/nightly-uuid)

src/lib.rs
```diff
@@ -714,10 +716,16 @@ impl Uuid {
     ///
     /// let uuid = Uuid::from_uuid_bytes(bytes);
     /// ```
+    #[cfg(not(feature = "nightly"))]
     pub fn from_uuid_bytes(b: UuidBytes) -> Uuid {
         Uuid { bytes: b }
     }
 
+    #[cfg(feature = "nightly")]
```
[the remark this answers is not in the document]
mokeyish: @kinggoesgaming ```#[cfg(nightly)]``` didn't working on  [here](https://play.rust-lang.org/?gist=5a15b198e0472b46383e47a24ca01259&version=nightly&mode=debug)

Cargo.toml
Dylan-DPC: based on the discussion in #220 we can go back to this. so this is fine

src/lib.rs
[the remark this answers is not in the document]
Dylan-DPC: based on the discussion in #220 we can go back to this. so this is fine

PR #221, opened 2018-05-09
Fast travis

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [ ] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
* Make travis builds faster
* Reduce the number of build environments to 9, was 14

# Motivation
Travis is bound to get slower over time, especially with some more feature sets coming in `breaking` branch. A speedup was in order

# Tests
General comparisons of the build times

# Related Issue(s)
N/A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
     before_script:
     - rustup component add rustfmt-preview
     script:
-    - cargo fmt --all -- --write-mode=diff
+    - cargo fmt --all -- --write-mode=check
     - cargo test --features "serde std v1 v3 v4 v5"
     - cargo bench --features "serde std v1 v3 v4 v5"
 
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
I feel it's better to test each feature in isolation as we could one feature affecting the other feature

KodrAus:
I think we should be safe testing different features in the same run. The tradeoff I see is in identifying the cause of build failures, but that shouldn't be too much harder than it is now.

I'm not sure about the OSX build though. Travis regularly gets large OSX backlogs that can stall builds for several hours, so I'd avoid it unless we've got some platform specific behaviour to test.

Sans OSX I think this is a worthwhile change :+1:

kinggoesgaming:
I could switch to 2 OS X builds for 1.18 and nightly?

kinggoesgaming:
this should be ready to go

Dylan-DPC:
bors: r+

Issue #239, opened 2018-05-21
update serde and serde_test


kinggoesgaming:
# Description
New releases of both crates are available

Dylan-DPC:
We don't have to do it. We are specifying "1.0.16" for serde which is equivalent to "^1.0.16" so it will cover all the latest versions of the crate.

kinggoesgaming:
i am particularly wanting it for `u128` support that serde included in `1.0.56`

PR #240, opened 2018-05-21
update serde and serde_test crates

kinggoesgaming:
**I'm submitting a ...**
  - [ ] bug fix
  - [ ] feature enhancement
  - [ ] deprecation or removal
  - [ ] refactor

# Description
New versions available of both `serde` and `serde_test`

# Motivation
`u128` support in `1.0.56` `serde`

# Tests
tests pass

# Related Issue(s)
#239

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ all-features = true
 features = ["serde", "u128", "v1", "v3", "v4", "v5"]
 
 [dependencies]
-serde = { version = "1", optional = true, default-features = false }
+serde = { version = "1.0.56", optional = true, default-features = false }
 rand = { version = "0.4", optional = true }
 sha1 = { version = "0.6", optional = true }
 md5 = { version = "0.3", optional = true }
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
We can't use the general "1" as a dependency specifier here, since we are depending on a particular "1.0.56" version here for 128 (even though it is optional) 

Better to use "1.0.56"

Cargo.toml
```diff
@@ -27,7 +27,7 @@ all-features = true
 features = ["serde", "u128", "v1", "v3", "v4", "v5"]
 
 [dependencies]
-serde = { version = "1.0.16", optional = true, default-features = false }
+serde = { version = "1", optional = true, default-features = false }
```
Dylan-DPC: We can't use the general "1" as a dependency specifier here, since we are depending on a particular "1.0.56" version here for 128 (even though it is optional)

Better to use "1.0.56"

kinggoesgaming: done

Dylan-DPC:
bors: r+

Issue #276, opened 2018-07-03
Refactor ParseError


kinggoesgaming:
Current `ParseError` assumes that only `Hyphenated`, `Simple` and `Urn` + `SIV 128` exist out in the universe. We need to address this so people can have their own parser for other formats (like URL blob uuid) and can provide their own context for the error.

PR #278, opened 2018-07-10
add UuidLength enum

kinggoesgaming:

# Description
Adds a `UuidLength` utility enum. Useful for parsing. The location is not final and may be moved around as we start using in the crate.

This is currently not used in the crate, however, future PRs are expected to rectify this.

# Motivation
Current parsing mechanism assumes only `Hyphenated`, `Simple`, `Urn` and `SIV` uuids exist. However this not true. This will help make parsers, especially external parsers for other formats 

# Tests
No tests currently. However, when other parts use this enum they can update their tests to use this

# Related Issue(s)
#276

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,6 +142,7 @@ use core::{fmt, str};
 pub mod adapter;
 pub mod ns;
 pub mod prelude;
+pub mod util;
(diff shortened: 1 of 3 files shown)


Dylan-DPC:
bors: r+

PR #158, opened 2018-03-02
adds optional slog::Value impl

jonathanstrong:
This implements `slog::Value` behind a feature-gate, allowing the use of `Uuid` in a key-value pair. Includes a test and bench.

PR #108, opened 2017-12-25
Rename `use_std` feature to `std`

whitequark:
The `use_std` feature remains for backwards compatibility.

Fixes #107.

PR #280, opened 2018-07-13
Remove deprecated "use-std"

Dylan-DPC:
**I'm submitting a(n)** removal

# Description
feature `use-std` was renamed earlier in #108 and brought back and marked deprecated in #158. Removing it now

# Related Issue(s)
#108, #158

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,11 +64,6 @@ v3 = ["md5", "rand"]
 v4 = ["rand"]
 v5 = ["sha1", "rand"]
 
-# deprecated
-#------------------------
-# Use `"std"` instead
-use_std = ["std"]
-
 # since rust 1.26.0
 u128 = ["byteorder"]
 


kinggoesgaming:
I am not really comfortable with removing it quite yet. I think we should reserve this for `0.8` and aggressively advertise the removal in `0.7.0`


@KodrAus opinions?

Dylan-DPC:
> aggressively advertise the removal in 0.7.0

Deprecating something is aggressively advertising people to remove it.

kinggoesgaming:
I am not strictly against this... but would love some opinion from people who use this library (especially those who provide stability guarantees like `1.0+`)

Dylan-DPC:
The users know that 0.7 is an breaking change, so they will expect stuff to break. Also most people use `"0.6"` as their dependency selector (if they use anything else such as `*`then there are multiple things which will already break their API), so it is not that this change will break their API overnight, and they will not just change the version from 0.6 to 0.7 without going through the list of things that have changed (which already consists of multiple API changes in the crate).

PR #295, opened 2018-08-14
make the current 2 parser functions crate public

kinggoesgaming:

# Description
Make `fn`s in `::parser` module crate private

# Motivation

These are considered experimental, so exposing them is not the
best idea


Dylan-DPC:
bors: r+

PR #301, opened 2018-08-20
refactored code to use const_ft

Dylan-DPC:

# Description
Adds const_ft(0.1.2) as a dependency
Refactors code to use the dependency

# Motivation
Code looks cleaner and avoids const(feature) & const(not-feature) repetition

# Tests
All tests pass, none added

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,9 @@ version = "0.6"
 optional = true
 version = "2"
 
+[dependencies.const-ft]
+version = "0.1.2"
+
 [dev-dependencies.serde_test]
 version = "1.0.56"
 
@@ -70,6 +73,6 @@ u128 = ["byteorder"]
 # nightly rust
 #------------------------
 # Allow using `const fn`s
-const_fn = ["nightly"]
+const_fn = ["nightly", "const-ft/const_fn"]
 # Nightly marker feature gate
 nightly = []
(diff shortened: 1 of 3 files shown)


KodrAus:
Hmm, I think we should be really mindful about introducing new dependencies. Do we need to use an external crate for const functions? Using a macro seems nice, but it could easily just live here in `uuid`.

kinggoesgaming:
As there some problems with the macro in itself, I think its best to not depend on it either way until they are polished out

Dylan-DPC:
@kodrAus it doesn't make much of a difference since it's my own crate. Though currently there are some issues with the macro so might postpone this

KodrAus:
@Dylan-DPC there are a few differences worth considering. The first is compile times where additional dependencies mean more units of compilation, more metadata, more network, so they'll regress slightly. The second is in mental overhead, where there's an additional non-local piece of infrastructure new contributors need to understand in order to contribute.

I'm not suggesting those are hard blockers, but we should definitely consider them.

Dylan-DPC:
@KodrAus I understand your point and it was a reason why i wasn't so keen on `cfg_if`. One of the main reasons of creating a new crate is that it wasn't related to uuids and can be something of greater use to other projects as well.

 Turns out there is still some more stuff to be done with the macros so closing the PR now and we can reconsider it later.

PR #303, opened 2018-08-23
 Optimize formatting by 80-90% by avoiding write!.

huonw:
**I'm submitting a(n)** refactor.

# Description

Using a temporary byte buffer along with manual decomposition of the
UUID means the code compiles down to a nice loop, instead of doing
multiple virtual function calls as it does with `write!`.

# Motivation

The code was unnecessarily slow.

# Tests

All the existing formatting tests pass, and I introduced a benchmark. Here's the output of `cargo benchcmp` on before/after results for that benchmark:

    name          before.txt ns/iter  after.txt ns/iter  diff ns/iter   diff %  speedup
    bench_hyphen  270                 49                         -221  -81.85%   x 5.51
    bench_simple  418                 40                         -378  -90.43%  x 10.45
    bench_urn     281                 56                         -225  -80.07%   x 5.02


KodrAus:
Looks good to me! Thanks @huonw

bors r+

PR #311, opened 2018-09-05
Add encode_{lower,upper} to formatting adapters for core::fmt-less rendering.

### LINKED ISSUE #196: Faster conversion to String
Hello, I had a need for much faster printing of UUIDs, and I created a library [uuidtoa](https://github.com/jwilm/uuidtoa) to prototype and use immediately in another project. I would like to send a PR to include the improved write speed here, but a couple of questions first.

1. Is the uuid project interested?
2. Not going through the format! machinery is a requirement for me. Would you be open to adding a public API like what's used in uuidtoa _in addition_ to the `fmt::Display` way of printing currently included in uuid?

Thanks!

huonw:

# Description

Add encode_{lower,upper} to formatting adapters for core::fmt-less rendering.

# Motivation

These are noticably faster (20ns, ~50%) as they avoid all the virtual
call overhead of core::fmt/write! (etc.). This also improves the
performance of (human-readable) serde encoding, since that previously
went through core::fmt.

# Tests

Benchmark results both with LTO and without including a comparison
against crates.io/crates/uuidtoa:

```
test lto_bench_encode_hyphen ... bench:          16 ns/iter (+/- 1)
test lto_bench_encode_simple ... bench:          14 ns/iter (+/- 1)
test lto_bench_encode_urn    ... bench:          18 ns/iter (+/- 2)
test lto_bench_hyphen        ... bench:          36 ns/iter (+/- 1)
test lto_bench_simple        ... bench:          33 ns/iter (+/- 7)
test lto_bench_urn           ... bench:          40 ns/iter (+/- 4)
test lto_bench_uuidtoa       ... bench:          13 ns/iter (+/- 3)

test nolto_bench_encode_hyphen ... bench:          18 ns/iter (+/- 3)
test nolto_bench_encode_simple ... bench:          19 ns/iter (+/- 3)
test nolto_bench_encode_urn    ... bench:          21 ns/iter (+/- 2)
test nolto_bench_hyphen        ... bench:          41 ns/iter (+/- 11)
test nolto_bench_simple        ... bench:          39 ns/iter (+/- 3)
test nolto_bench_urn           ... bench:          48 ns/iter (+/- 6)
test nolto_bench_uuidtoa       ... bench:          41 ns/iter (+/- 5)
```

This is now essentially as fast as uuidtoa at its fastest (and
significantly faster than it, without LTO), and the remaining
difference is likely the use of the checked `str::from_utf8_mut`.

Benchcmp for the existing benchmarks using the existing APIs (only the
ones that changed):

```
 name                        before.txt ns/iter  after.txt ns/iter  diff ns/iter   diff %  speedup
 lto_bench_hyphen            43                  36                           -7  -16.28%   x 1.19
 lto_bench_json_encode       59 (644 MB/s)       35 (1085 MB/s)              -24  -40.68%   x 1.69
 lto_bench_urn               45                  40                           -5  -11.11%   x 1.12
 nolto_bench_hyphen          48                  41                           -7  -14.58%   x 1.17
 nolto_bench_json_encode     73 (520 MB/s)       45 (844 MB/s)               -28  -38.36%   x 1.62
 nolto_bench_urn             59                  48                          -11  -18.64%   x 1.23
```

That is, JSON encoding takes 40% less time, and even core::fmt
formatting for the formats that use hyphens got slightly faster too.

# Related Issue(s)

Closes #196.

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -13,6 +13,7 @@
 //!
 //! [`Uuid`]: ../struct.Uuid.html
 
+use core::str;
 use prelude::*;
 
 mod core_support;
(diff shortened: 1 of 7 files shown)


src/adapter/mod.rs
```diff
@@ -187,6 +188,59 @@ impl Uuid {
...
+                buffer[out_idx + 1] = hex[(b & 0b1111) as usize];
+            }
+
+            if group != 4 && hyphens {
+                buffer[HYPHEN_POSITIONS[group]] = b'-';
+            }
+        }
+    }
+
+    str::from_utf8_mut(&mut full_buffer[..start + len]).unwrap()
```
KodrAus: It might be nicer to use `expect` with a message here indicating that this shouldn't happen rather than `unwrap`.

src/adapter/mod.rs
```diff
@@ -187,6 +188,59 @@ impl Uuid {
...
+];
+/// The segments of a UUID's [u8; 16] corresponding to each group.
+const BYTE_POSITIONS: [usize; 6] = [0, 4, 6, 8, 10, 16];
+/// The locations that hyphens are written into the buffer, after each
+/// group.
+const HYPHEN_POSITIONS: [usize; 4] = [8, 13, 18, 23];
+
+fn encode<'a>(
+    full_buffer: &'a mut [u8],
+    start: usize,
```
KodrAus: Is there a reason to pass a start index in rather than expect `full_buffer` to have been sliced?

huonw: Yeah, for the URN case: there's a prefix that gets added in front of what `encode` does, but I want to have that prefix included in the return value, and thus also in the `from_utf8_mut` call in this function.

KodrAus: Ah gotcha :+1:

src/serde_support.rs
```diff
@@ -19,7 +19,8 @@ impl Serialize for Uuid {
         serializer: S,
     ) -> Result<S::Ok, S::Error> {
         if serializer.is_human_readable() {
-            serializer.collect_str(&self.to_hyphenated())
```
kinggoesgaming: just a question: was `collect_str` the wrong function to use? curious..

huonw: I don't think so, it's much more direct and less error prone than manually using the `Display` impls to construct a `str`. This new APIs is only a little bit more complicated to use with `serialize_str` than `collect_str`, so I think it makes sense to switch.

src/adapter/mod.rs
huonw: Done.

src/adapter/mod.rs
huonw: I added a doc comment to `encode` that explains this.

kinggoesgaming:
@KodrAus we are waiting until `0.7.0‘ release before merging this

Dylan-DPC:
bors: r+

PR #317, opened 2018-09-13
make encode_ functions pub(crate) for now

kinggoesgaming:
**I'm submitting a(n)** other

# Description
Includes:
* `uuid::Uuid::encode_buffer() -> [u8; adapter::Urn::LENGTH]`
* &lt;adapter>::{encode_lower, encode_upper}

# Motivation
The function signature _may_ change but we don't want a breaking change after `0.7.1` if we can help it.

# Tests
Tests pass :)

# Related Issue(s)
#311


huonw:
I'm curious: why did the doc comments and tests need to be commented for this change? `pub(crate)` should affect those?

kinggoesgaming:
For some reason tests and do tests were complaining about private fns

Issue #277, opened 2018-07-06
Little-Endian byte order support


awakecoding:
Microsoft uses little-endian byte order representation for its GUID structure in memory, like this:

``
typedef struct _GUID
{
    UINT32 Data1;
    UINT16 Data2;
    UINT16 Data3;
    BYTE Data4[8];
} GUID, UUID, *PGUID, *LPGUID, *LPCGUID;
``

Would it be possible to add the option of reading from byte arrays in little endian order, and writing to byte arrays in little endian order? The current code apparently assumes big-endian ordering 100% of the time, making it harder to interoperate in specific cases.

For instance, from_bytes_le and as_bytes_le could be special functions to read/write to/from bytes in little-endian order.

Dylan-DPC:
@awakecoding Thanks for proposing this. You can go ahead and write a PR if you think you can work on it else one of us will do it.

kinggoesgaming:
If you do open a PR, please target `breaking` branch. (we will soon be merging it into `master` but for now we still have active development of `breaking`)

awakecoding:
@Dylan-DPC @kinggoesgaming I don't know if I should try to do it myself or let you decide how it should be done, as I am unsure how you would like this to be implemented in a proper manner. Do you have a preference for the new API? I don't mind if the internal representation remains in big endian, as long as we can easily import/export to/from little-endian order bytes.

BurntSushi:
@kinggoesgaming Note that `byteorder`'s MSRV is 1.12 currently.

Speaking from experience, I'd probably prefer solutions that don't incur additional public dependencies.

awakecoding:
I agree that introducing a new public dependency is probably not a good idea

kinggoesgaming:
@BurntSushi we already depend on `byteorder` for `u128` to `Uuid` conversion and does mean minimum version of 1.26

BurntSushi:
Oh, okay. I guess I got confused. Sounded like MSRV and byteorder as a public dep were related. But if you are unconditionally using u128 internally, then yeah, your MSRV has nothing to do with byteorder really.

kinggoesgaming:
> Related - it would be handy to support the winapi-rs crate by implementing `std::convert::From<GUID>`.

opened a new issue #333

KodrAus:
> The current code apparently assumes big-endian ordering 100% of the time, making it harder to interoperate in specific cases.

[...]

PR #330, opened 2018-10-01
Add little-endian functions

Veykril:
**I'm submitting a feature**

# Description
This PR adds little-endian constructors as well as a `to_bytes_le` and a `to_fields_le` method. I've also renamed the previous bytes methods and appended the be suffix for consistency while adding the previous function names as wrapper methods around those to not break things. All of this is personal preference of course so far so please point out changes you want to have with this.

Edit: I forgot to mention that the le methods use the `to_*` naming conventions instead of `as_*` since they aren't a free conversion and dont return a `&Bytes`

# Motivation
Issue #277.

# Tests
Tested via doc-tests

# Related Issue(s)
#277

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -752,11 +752,11 @@ impl Uuid {
     /// original `Uuid` back.
     ///
     /// * The first field value represents the first group of (eight) hex
-    ///   digits, taken as a little-endian `u32` value.  For V1 UUIDs, this field
-    ///   represents the low 32 bits of the timestamp.
+    ///   digits, taken as a little-endian `u32` value.  For V1 UUIDs, this
+    ///   field represents the low 32 bits of the timestamp.
     /// * The second field value represents the second group of (four) hex
-    ///   digits, taken as a little-endian `u16` value.  For V1 UUIDs, this field
-    ///   represents the middle 16 bits of the timestamp.
+    ///   digits, taken as a little-endian `u16` value.  For V1 UUIDs, this
+    ///   field represents the middle 16 bits of the timestamp.
     /// * The third field value represents the third group of (four) hex
     ///   digits, taken as a little-endian `u16` value.  The 4 most significant
     ///   bits give the UUID version, and for V1 UUIDs, the last 12 bits
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -578,47 +578,7 @@ impl Uuid {
     ///
     /// let uuid = Uuid::from_bytes_le(bytes);
     /// ```
-    #[cfg(not(feature = "const_fn"))]
     pub fn from_bytes_le(mut bytes: Bytes) -> Uuid {
-        Self::be_to_le_bytes(&mut bytes);
-        Self::from_bytes_be(bytes)
-    }
-
-    /// Creates a `Uuid` using the supplied little-endian bytes.
-    ///
-    /// # Examples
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -539,13 +538,105 @@ impl Uuid {
...
+    /// let uuid = Uuid::from_bytes_le(bytes);
+    /// ```
+    #[cfg(feature = "const_fn")]
+    pub const fn from_bytes_le(mut bytes: Bytes) -> Uuid {
+        Self::be_to_le_bytes(&mut bytes);
+        Self::from_bytes_be(bytes)
+    }
+
+    /// Creates a `Uuid` using the supplied big-endian bytes.
+    /// This method is wraps [`from_bytes_be`]: #method.from_bytes_be
```
Dylan-DPC: // This methods wraps

src/lib.rs
```diff
@@ -539,13 +538,105 @@ impl Uuid {
...
+    #[cfg(not(feature = "const_fn"))]
+    pub fn from_bytes(bytes: Bytes) -> Uuid {
+        Self::from_bytes_be(bytes)
+    }
+
+    /// Creates a `Uuid` using the supplied big-endian bytes.
+    /// This method is wraps [`from_bytes_be`]: #method.from_bytes_be
```
Dylan-DPC: same as prev comment

src/lib.rs
```diff
@@ -539,13 +538,105 @@ impl Uuid {
...
+    /// use uuid::Bytes;
+    /// use uuid::Uuid;
+    ///
+    /// let bytes: Bytes = [4, 54, 67, 12, 43, 2, 98, 76]; // doesn't compile
+    ///
+    /// let uuid = Uuid::from_bytes_le(bytes);
+    /// ```
+    #[cfg(not(feature = "const_fn"))]
+    pub fn from_bytes_le(mut bytes: Bytes) -> Uuid {
+        Self::be_to_le_bytes(&mut bytes);
```
Dylan-DPC: if we are expecting little-endian bytes then why are we converting it from big-endian to little-endian?

Veykril: Because we still keep the internal big-endian representation

Dylan-DPC: Ah that's fine. It was a bit odd when I read the code. :+1:

Dylan-DPC:
@Veykril once you are done, run `cargo fmt` using the latest nightly.

Dylan-DPC:
bors: r+

Veykril:
How exactly does it mess with serde? The internal representation didn't change so it shouldn't have any impact with serialization, or am I missing something?

kinggoesgaming:
The user might be using something like `bincode` and expecting the LE read AND LE write... but serde ofc doesnt do that... its a BE write

kinggoesgaming:
@Veykril we are currently discussing this on gitter if you want to join in: http://gitter.im/uuid-rs/Lobby

Dylan-DPC:
So what we were discussing is: 

> I am thinking of introducing 2 newtypes that allow that... `UuidBe` and `UuidLe`*
or we can just have a From/Into impls...
the way i was thinking was use UuidBe and UuidLe as adapters instead
so conversion is done explicitly by the user
this way the internal layout of Uuid bytes is left as impl detail

PR #331, opened 2018-10-02
Fix #329

### LINKED ISSUE #329: Support serialization as `[u8; 16]` rather than as `&[u8]`
Currently, serialization and deserialization are performed as a slice rather than a fixed-sized array.  When using the `bincode` codec, this means that the raw data is prefixed with a redundant length tag (`16usize`), which is unsuitable for loading binary datastructures containing UUIDs.  It is also unexpected that a UUID is serialized in a different way than `[u8; 16]`, since it essentially _is_ an `[u8; 16]`.

I would like to be able to serialize a UUID as if it was a `[u8; 16]` in non-human-readable codecs.  This could be protected by a new feature flag to avoid breaking compatibility with old serialized data.

**Is it blocking?**
Not really, since we stopped using this crate because of this issue.  While it would be possible to implement a different serialization for the `UUID` type, it was easier for us to use a plain `[u8;16]` instead.

**Describe alternatives you've considered**
I looked into writing my own serialization and deserialization functions.  It would be nice if this was supported out of the box, though.

**Additional context**
N/A

**Other**
N/A

Redrield:
**I'm submitting a** feature

# Description
Adds the ability to serialize a Uuid as a [u8; 16]. Functionality is behind the `dense_serde` feature. the `serde` feature continues to serialize as a &[u8]. Both features serialize strings identically

# Tests
A new test module for the feature was added, mirroring the tests of the `serde` feature (with modifications to expected tokens where necessary)

# Related Issue(s)
#329


KodrAus:
Thanks @Redrield!

From the discussion on #329, using a feature flag here doesn't actually save us from potential breakage. It looks like we want to keep the output of `serde` the same for anyone that's currently depending on it.

One way we could do this in a backwards compatible way that users properly opt into would be to convert your new `Serialize` and `Deserialize` impls into free functions, that can be used with the `#[serde(serialize_with)]` and `#[serde(deserialize_with)]` attributes.

kinggoesgaming:
@KodrAus newtypes?

smarnach:
@kinggoesgaming I think a newtype isn't a natural solution here.  It's really the same type, just serialized in a different way, and it may even depend on the context what serialization to use.  I like the suggestion to use free functions for use with serde attributes, since it makes the intention of client code immediately obvious.

Redrield:
I'm thinking add a module that can be used with `#[serde(with = "uuid::dense_serde")]`. (I'm not sure about the name, just an example)

KodrAus:
What do you think of `uuid::serde::compact`? That way we've got a `serde` module too that any other `serde` specific stuff can live in?

kinggoesgaming:
**note** `uuid::serde` namespace can't be used because it conflicts with `serde` crate

KodrAus:
Ok, so we could do `uuid::adapter::compact`, or `uuid::adapter::dense` instead ?

Redrield:
Rewritten to use serde attributes for dense serialization under uuid::adapter::compact

src/adapter/compact.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::fmt;
+
+/// Serializer for a [`Uuid`] into a `[u8; 16]`
+///
+/// [`Uuid`]: ../../struct.Uuid.html
+pub fn serialize<S: Serializer>(
+    u: &Uuid,
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    println!("sereeeeeeeeealize");
```
Dylan-DPC: uhmm?

src/adapter/compact.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/// Deserializer from a `[u8; 16]` into a [`Uuid`]
+///
+/// [`Uuid`]: ../../struct.Uuid.html
+pub fn deserialize<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<Uuid, D::Error> {
+    println!("desereeeeeeeeialize");
+    struct DenseUuidBytesVisitor;
```
Dylan-DPC: do we need the struct inside of the function? Generally not a fan of this.

src/adapter/compact.rs
Redrield: Debug messages, i'll take those out, whoops!

src/adapter/compact.rs
Redrield: I was following in the pattern of the String visitors in the (De)Serialize impls in serde_support.rs, it can be moved if you prefer

Dylan-DPC: It is fine. Just something I don't like personally but i'm okay with it if it keeps it consistent with other impls in the crate

src/adapter/compact.rs
Redrield: wdym? Removed them in b1b2efe

src/adapter/compact.rs
kinggoesgaming: Huuuuuuggggeee necro...


It just keeps it very much contained and prevents any accidental use outside the function

Issue #171, opened 2018-03-17
Builder pattern


Dylan-DPC:
# Description

We're starting to get a lot of potential sources for Uuids and different mutations to apply to them. Something we could look at is adding a builder that can be used to safely mutate a Uuid before handing it back:

```rust
// Builder: !Copy
pub struct Builder {
    inner: Uuid,
}

impl Builder {
    fn from_bytes(b: &[u8]) -> Result<Self, ParseError>;
    fn from_uuid_bytes(b: [u8; 16]) -> Self;
    fn nil() -> Self;

    fn set_variant(&mut self, v: UuidVariant) -> &mut Self;
    fn set_version(&mut self, v: UuidVersion) -> &mut Self;

    fn build(&mut self) -> Uuid;
}
let uuid = Builder::from_uuid_bytes(b)
    .set_variant(UuidVariant::RFC4122)
    .set_version(UuidVersion::Random)
    .build();
The builder would probably require some more thought to complement our current set of constructors on Uuid.
```

Related #170

kinggoesgaming:
closed in #332

PR #332, opened 2018-10-02
Add Uuid Builder

Veykril:
**I'm submitting a feature**

# Description
This PR implements a basic UuidBuilder that takes itself by `&mut` instead of `self` as shown in the issue #171 description.

# Motivation
Issue #171

# Tests
Through Doctests

# Related Issue(s)
#171

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,0 +1,241 @@
+use super::Bytes;
+use super::BytesError;
+use super::Uuid;
+use super::Variant;
+use super::Version;
+
+/// A builder struct for creating a [`Uuid`]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct UuidBuilder {
+    uuid: Uuid,
+}
+
+impl UuidBuilder {
+    /// Creates a `UuidBuilder` using the supplied big-endian bytes.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// use uuid::Bytes;
+    /// use uuid::UuidBuilder;
+    ///
+    /// let bytes: Bytes = [
+    ///     70, 235, 208, 238, 14, 109, 67, 201, 185, 13, 204, 195, 90, 145, 63,
+    ///     62,
+    /// ];
+    ///
+    /// let builder = UuidBuilder::from_bytes_be(bytes);
+    /// let uuid = builder.build().to_hyphenated().to_string();
+    ///
+    /// let expected_uuid = String::from("46ebd0ee-0e6d-43c9-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    /// An incorrect number of bytes:
+    ///
+    /// ```compile_fail
+    /// use uuid::Bytes;
+    /// use uuid::UuidBuilder;
+    ///
+    /// let bytes: Bytes = [4, 54, 67, 12, 43, 2, 98, 76]; // doesn't compile
+    ///
+    /// let uuid = UuidBuilder::from_bytes_le(bytes);
+    /// ```
+    pub fn from_bytes_be(b: Bytes) -> Self {
+        UuidBuilder {
+            uuid: Uuid::from_bytes_be(b),
+        }
+    }
+
+    /// Creates a `UuidBuilder` using the supplied little-endian bytes.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// use uuid::Bytes;
+    /// use uuid::UuidBuilder;
+    ///
+    /// let bytes: Bytes = [
+    ///     70, 235, 208, 238, 14, 109, 67, 201, 185, 13, 204, 195, 90, 145, 63,
+    ///     62,
+    /// ];
+    ///
+    /// let builder = UuidBuilder::from_bytes_le(bytes);
+    /// let uuid = builder.build().to_hyphenated().to_string();
+    ///
+    /// let expected_uuid = String::from("eed0eb46-6d0e-c943-b90d-ccc35a913f3e");
+    ///
+    /// assert_eq!(expected_uuid, uuid);
+    /// ```
+    ///
+    /// An incorrect number of bytes:
+    ///
+    /// ```compile_fail
+    /// use uuid::Bytes;
+    /// use uuid::UuidBuilder;
+    ///
+    /// let bytes: Bytes = [4, 54, 67, 12, 43, 2, 98, 76]; // doesn't compile
+    ///
+    /// let uuid = UuidBuilder::from_bytes_le(bytes);
+    /// ```
+    pub fn from_bytes_le(b: Bytes) -> Self {
+        UuidBuilder {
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -0,0 +1,241 @@
+use super::Bytes;
+use super::BytesError;
+use super::Uuid;
+use super::Variant;
+use super::Version;
+
+/// A builder struct for creating a [`Uuid`]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
KodrAus: Let's remove the `Copy` implementation here, because the behaviour of `Copy` types and `&mut self` methods can be surprising.

KodrAus:
Hi @Veykril :wave:

I'm on mobile at the moment so just left a little comment from a quick scan.

I'd be tempted to mark the `from_random_bytes` method as deprecated now and suggest using this builder type.

src/builder.rs
Veykril: Oh interesting, I didn't know that this might cause weird behaviour. Will remove it then, this builder doesn't consume itself anyways so there is no need to copy it anyways.

Veykril:
Should I deprecate it with `#[deprecated(since="0.7.2", note="please use the `UuidBuilder` instead")]` then?

Dylan-DPC:
I don't think it's a good idea to deprecate we are just giving the users to
use a builder pattern if they want to.

src/builder.rs
Veykril: Actually removing the copy implementation will trigger the `missing_copy_implementations` lint.

KodrAus:
@Dylan-DPC I think `from_random_bytes` was really only added specifically to call some of the mutating methods we now have on this builder without exposing them on `Uuid`, which is `Copy`. I don't think we need to hang on to APIs indefinitely when new approaches emerge, otherwise we could risk winding up with a large API surface area that's difficult to navigate. Pushing users towards the `Builder` for any non-standard initialisation seems like it would be helpful for users trying to find their way around.

Dylan-DPC:
@KodrAus oops yeah just checked the function again. That's fine by me :+1:

src/builder.rs
KodrAus: > Oh interesting, I didn't know that this might cause weird behaviour.

Yeh, the wierdness is that a type that implements `Copy` can be implicitly copied around, so when you're passing the value around and calling mutating methods you can end up mutating a copy by accident.

Veykril:
Alright, I made the proposed changes. Is the version I put into the attribute correct for this?

Dylan-DPC:
yep @Veykril that's fine

src/builder.rs
Veykril: Ah yes, that makes never though of it like that. That is good thing to keep in mind.

src/builder.rs
```diff
@@ -0,0 +1,238 @@
+use super::Bytes;
+use super::BytesError;
+use super::Uuid;
```
kinggoesgaming: use `uuid::prelude::*` where possible

src/lib.rs
```diff
@@ -137,6 +137,9 @@ extern crate sha1;
 #[cfg_attr(test, macro_use)]
 extern crate slog;
 
+mod builder;
+pub use builder::UuidBuilder;
```
kinggoesgaming: make the `builder` module `pub` and instead of the reexport here, reexport in `prelude` module

src/lib.rs
```diff
@@ -619,6 +622,10 @@ impl Uuid {
     ///
     /// assert_eq!(expected_uuid, uuid);
     /// ```
+    #[deprecated(
+        since = "0.7.2",
+        note = "please use the `UuidBuilder` instead"
```
kinggoesgaming: `note = "use `uuid::builder::UuidBuilder` instead"`

src/builder.rs
```diff
@@ -0,0 +1,238 @@
...
+use super::Variant;
+use super::Version;
+
+/// A builder struct for creating a [`Uuid`]
+#[allow(missing_copy_implementations)]
+#[derive(Debug)]
+pub struct UuidBuilder {
```
kinggoesgaming: maybe make this a `newtype` tuple? Just to make it consistent with the rest of the crate

src/builder.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// use uuid::UuidBuilder;
+    ///
+    /// let uuid = UuidBuilder::nil().build();
+    ///
+    /// assert_eq!(
+    ///     uuid.to_hyphenated().to_string(),
+    ///     "00000000-0000-0000-0000-000000000000"
+    /// );
+    /// ```
+    pub fn build(&self) -> Uuid {
```
kinggoesgaming: `pub fn build(&self) -> Uuid` -> `pub fn build(self) -> Uuid`

src/builder.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// use uuid::prelude::*;
+    /// use uuid::UuidBuilder;
+    ///
+    /// let bytes = [4, 54, 67, 12, 43, 2, 98, 76];
+    ///
+    /// let builder = UuidBuilder::from_slice(&bytes);
+    ///
+    /// assert!(builder.is_err());
+    /// ```
+    pub fn from_slice(b: &[u8]) -> Result<Self, BytesError> {
```
kinggoesgaming: maybe have two be/le `from_slice` functions instead?

src/builder.rs
KodrAus: We could make this method take `&mut self`, so that we have the future option to move the internal uuid out instead of copying it.

src/lib.rs
KodrAus: Something about `uuid::builder::UuidBuilder` doesn't sit right with me... We could just go with `uuid::Builder`?

Dylan-DPC: yeah its not like we are going to have several builders so `uuid::Builder` should be  fine

Veykril: So should I reexport the `Builder` via `pub use builder::Builder;` in the `lib.rs` as well as reexporting it through the prelude?

src/builder.rs
Veykril: I used `&self` since that was what was given in the issue example, normally you'd consume the builder when building but since our internal type is copy both works here I'd say. What should I pick?

KodrAus: I would go with `&mut self` here, it's consistent with the other mutating methods we've added, and lets us do some extra work in the `build` method if we need it. Using `&mut self` over `self` means that we can chain the method calls in a single statement like this:

```rust
let uuid = Builder::from_slice(bytes).set_version(..).build();
```

If the `build` method took `self` by value then we wouldn't be able to chain it with `set_version`.

Veykril:
I think I changed everything that was requested now

kinggoesgaming:
@Veykril can fix the build failures and then its good to merge

kinggoesgaming:
I dont know what that error message is :confused: 

@Dylan-DPC @KodrAus  ideas?

Veykril:
Oh I think Rust 1.22 doesn't support nested import groups, yeah that feature was added in 1.25.

Dylan-DPC:
bors: r+

PR #336, opened 2018-10-06
Add support for converting from winapi types

### LINKED ISSUE #333: support the winapi-rs crate by implementing `std::convert::From<GUID>`.
_Originally posted by @lzybkr in https://github.com/uuid-rs/uuid/issues/277#issuecomment-426824514_

Redrield:

# Description
Add a module `guid_support` that contains methods on Uuid for converting to and from winapi::shared::guiddef::GUID. Along with this an optional dependency on winapi is added.

# Motivation
#333 

# Tests
Not yet, waiting for a VM (will update once I have run tests)

# Related Issue(s)
closes #333


src/lib.rs
```diff
@@ -160,6 +163,8 @@ mod std_support;
 mod test_util;
 #[cfg(feature = "u128")]
 mod u128_support;
+#[cfg(feature = "winapi")]
+mod guid_support;
```
kinggoesgaming: `winapi_support`

Cargo.toml
```diff
@@ -26,6 +26,10 @@ all-features = true
 [package.metadata.playground]
 features = ["serde", "u128", "v1", "v3", "v4", "v5"]
 
+[dependencies.winapi]
+version = "0.3.6"
```
kinggoesgaming: `0.3`

src/guid_support.rs
```diff
@@ -0,0 +1,32 @@
+use prelude::*;
+
+use BytesError;
+
+use winapi::shared::guiddef::GUID;
```
kinggoesgaming: `use winapi::shared::guiddef`

src/guid_support.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl Uuid {
+    /// Attempts to create a [`Uuid`] from a winapi `GUID`
+    ///
+    /// [`Uuid`]: ../struct.Uuid.html
+    fn from_guid(guid: GUID) -> Result<Uuid, BytesError> {
```
kinggoesgaming: `guid: guiddef::GUID`

src/guid_support.rs
```diff
@@ -0,0 +1,32 @@
...
+            &(guid.Data4 as [u8; 8]),
+        )
+    }
+
+    /// Converts a [`Uuid`] into a winapi `GUID`
+    ///
+    /// [`Uuid`]: ../struct.Uuid.html
+    fn to_guid(&self) -> GUID {
+        let (data1, data2, data3, data4) = self.to_fields_le();
+        GUID {
```
kinggoesgaming: `guiddef::GUID`

src/winapi_support.rs
```diff
@@ -0,0 +1,32 @@
...
+            guid.Data3 as u16,
+            &(guid.Data4 as [u8; 8]),
+        )
+    }
+
+    /// Converts a [`Uuid`] into a winapi `GUID`
+    ///
+    /// [`Uuid`]: ../struct.Uuid.html
+    fn to_guid(&self) -> guiddef::GUID {
+        let (data1, data2, data3, data4) = self.to_fields_le();
```
kinggoesgaming: Document assumption that the `Uuid` is LE

kinggoesgaming:
@Redrield thanks for working on this.

Get the assumption documented and it will be ready for merge

src/winapi_support.rs
Redrield: I could have it use whichever endianness the `Uuid` is. `Uuid::to_fields_le` is the only method I found, and is actually the outlier, with the other methods being prefixed `as_`

Redrield:
Another thing that I could do, now I think of it is make `Uuid::to_guid_le` and `Uuid::to_guid_be` if wanted

KodrAus:
> Doc comments are changed, I'm still waiting for this VM so that I can add tests and such to verify functionality.

:+1: We'll need some roundtrip tests to make sure `Uuid`s converted into `Guid`s result in the same values, and that `Uuid` assigns the fields of the `Guid` in the same way as Windows APIs do.

KodrAus:
> Another thing that I could do, now I think of it is make Uuid::to_guid_le and Uuid::to_guid_be if wanted

Hmm, I don't think this would make a lot of sense here, would it? Since the fields of the `Guid` are primitive integers, they'll just use the platform endianness.

Redrield:
Got this code for a test
```
    #[test]
    fn test_conversion() {
        let uuid = Uuid::from_str("735d359d-4bc4-4e07-8c49-eb3e99a048dc").unwrap();
        let guid = uuid.to_guid();
        assert_eq!(Ok(uuid), Uuid::from_guid(guid));
    }
```

and it's failing, looks like from an endian mismatch. from_guid uses from_fields, so is that endian dependant?

KodrAus:
@Redrield I'm thinking the `from_fields_le` call is the problem here and we probably want to use `from_fields` instead.

Redrield:
There is no `from_fields_le` from what I can see. There's `from_bytes_le` but that takes a single unified array, not the individual fields

KodrAus:
Sorry, I meant `to_fields` rather than `to_fields_le` :)

Redrield:
That was indeed the problem, are there any other tests that are going to be wanted other than that one? I'll add them before I push the final code for merge

Dylan-DPC:
@Redrield one from guid to uuid will be helpful

Dylan-DPC:
bors: r+

PR #339, opened 2018-10-08
Cleanup the code

### LINKED ISSUE #282: Reenable clippy
We previously had `clippy` enabled, but had to disable it as it was **very** unstable and was breaking every day. 

However, now that it is available as a component, with the guarantee that nightly will not be available if it doesn't successfully compile `clippy`, we should now be able do clippy tests without that issue

kinggoesgaming:

# Description
Used clippy on stable channel for most features (except guid, since I don't have windows available)

# Motivation
Will be creating another PR that closes #282. But wanted to catch the current issues before doing that

# Tests
Current tests pass. No new tests needed

# Related Issue(s)
N/A

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -238,7 +238,7 @@ fn encode<'a>(
                 let b = bytes[idx];
                 let out_idx = hyphens_before + 2 * idx;
 
-                buffer[out_idx + 0] = hex[(b >> 4) as usize];
+                buffer[out_idx] = hex[(b >> 4) as usize];
                 buffer[out_idx + 1] = hex[(b & 0b1111) as usize];
             }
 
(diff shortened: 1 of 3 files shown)


Dylan-DPC:
bors: r+

src/adapter/mod.rs
```diff
@@ -234,11 +234,15 @@ fn encode<'a>(
             // number.
             let hyphens_before = if hyphens { group } else { 0 };
 
-            for idx in BYTE_POSITIONS[group]..BYTE_POSITIONS[group + 1] {
-                let b = bytes[idx];
+            for (idx, b) in bytes
```
KodrAus: @kinggoesgaming Curious to know whether this change results in any regression in our microbenchmarks.

kinggoesgaming: hmm.. hadnt considered that.. I will have a look

PR #340, opened 2018-10-15
[wip] start splitting const_fn and non const_fn stuff

kinggoesgaming:

# Description
split `const_fn` implementations into its own module

# Motivation
* `impl` blocks are crowded with `const fn` and non-`const fn` variants.
* I am working on expanding the functions that have `const fn` support (using `const_let`), which would make this situation even worse.

# Tests
current tests should pass. no additional tests.

# Related Issue(s)


kinggoesgaming:
This will no longer be needed as we move to `2018` edition

PR #341, opened 2018-10-15
reenable clippy

### LINKED ISSUE #282: Reenable clippy

kinggoesgaming:
**I'm submitting a(n)** other

# Description
Finally reenabled `clippy` :)

# Motivation
`clippy` is now provided as a component.

# Tests
Tests should pass.

# Related Issue(s)
closes #282

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,13 @@ matrix:
     - cargo bench --features "serde std v1 v3 v4 v5"
   - os: linux
     rust: nightly
+    before_script:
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
bors: r+

PR #345, opened 2018-11-05
[WIP] Always store bytes as big endian

KodrAus:

# Description

This PR reworks our endianness support so that we'll always store the uuid bytes using a specific endianness (rather than the current platform-dependent one). Fields are always treated as being in the platform's native endianness.

Right now I'm just working through how we can test both endiannesses in CI. The code doesn't actually work yet. Once that's all spiked up I'll update this description with more details.

# Related Issue(s)

For #277 

Following from #330


KodrAus:
Another way we could approach this is to always assume that fields given are already big endian and not mess with their layout at all. That would avoid breaking anyone with the current API so is probably the better way to go.

That means that for `Guid`s we'd have to assume the fields are little endian and flip them when converting to and from `Uuid`s. So the fields of a `Guid` converted into a `Uuid` will not have the same values.

We shouldn't need any endianness specific `bytes` methods, but instead would have:

- `from_fields`
- `from_fields_le`
- `to_fields`
- `to_fields_le`

Having alternative `*_be` methods would probably just confuse things even more.

Dylan-DPC:
bors: r+

PR #346, opened 2018-11-09
Use provided serde impls for compact serialization

KodrAus:

# Description

We don't actually need to provide our own implementation of serializing a 16-byte array because `serde` has one already we can lean on.


Dylan-DPC:
bors: r+

Issue #350, opened 2018-11-26
Uuid::new_v4() compiles and then panics for wasm32-unknown-unknown target


zrzka:
**Is your feature request related to a problem? Please describe.**

I do use `uuid` crate in my crate, which can be used as a pure Rust library or as an isomorphic NPM package (node & browser). The only feature I do use is random UUID (v4) generation.

It panics:

```
panicked at 'could not initialize thread_rng: No entropy sources available (permanently unavailable)', /Users/robertvojta/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.5.5/src/rngs/thread.rs:84:17
```

Example of what we do is in [this issue](https://github.com/balena-io-modules/balena-temen/issues/35). Our `uuidv4()` function [source code](https://github.com/balena-io-modules/balena-temen/blob/master/src/builtin/function/uuidv4.rs#L9-L11).

**Describe the solution you'd like**

Generate UUID v4 even when compiled for the `wasm32-unknown-unknown` target.

**Is it blocking?**

Kind of. Not a high priority.

**Describe alternatives you've considered**

Thinking about using `Math.random` in my function if the target is `wasm32-...`.

**Other**

Not sure if this is the `uuid` crate problem or it lies somehwere deeper inside this whole bleeding edge stuff.

Dylan-DPC:
Hi. Thanks for reporting this. 

Rand added wasm support to 0.6, while we are still using 0.5. (We have [a PR](https://github.com/uuid-rs/uuid/pull/348) to upgrade it to 0.6 but haven't merged it yet). 

You can test it by [depending](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) on the `dependabot/cargo/rand-0.6` branch instead of master.

zrzka:
@Dylan-DPC thanks, just tried ...

Browser

```
panicked at 'could not initialize thread_rng: No entropy sources available
  (permanently unavailable)', .../rand-0.6.1/src/rngs/thread.rs:82:17
```

Node

```
panicked at 'could not initialize thread_rng: No entropy sources available
  (permanently unavailable)', .../rand-0.6.1/src/rngs/thread.rs:82:17
```

Testing with:

```
uuid = { 
  features = ["v4"],
  git = "https://github.com/uuid-rs/uuid",
  branch = "dependabot/cargo/rand-0.6" 
}
```

zrzka:
[...]

PR #351, opened 2018-11-27
Feature/wasm

zrzka:
# Description

This PR adds two new features:

* `stdweb`
* `wasm-bindgen`

These features are kind of _passthrough_ features, because they do nothing in the `uuid` crate itself. They're just passed to the `rand` crate to make the `uuid` crate working for the `wasm32-unknown-unknown` target.

# Motivation

I'm unable to generate random UUID (v4) when this crate is compiled for the `wasm32-unknown-unknown` target.

# Tests

I just added these features ...

```
- cargo test --features "v3"
- cargo test --features "v3 stdweb"
- cargo test --features "v3 wasm-bindgen"
```

... for all `v3` & `v4` & `v5` (`rand` crate is used in all these features) to the `script` section in the `.travis.yml`.

Not sure if it makes sense, but it can demonstrate that it's buildable at least.

I don't think that more tests are required unless you'd like to bring the whole `wasm-bindgen`, `wasm-pack`, ... machinery here. And it has no sense to do it, because goal of this PR is not to publish `uuid-rs` NPM package, just add the ability to compile & use it from the `wasm32-unknown-unknown` target.

# Related Issue(s)

* #350
* [now() doesn't work](https://github.com/balena-io-modules/balena-temen/issues/37)

# Manual tests

My Cargo.toml:

```
[dependencies]
uuid = { features = ["v4"], git = "https://github.com/zrzka/uuid.git", branch = "feature/wasm" }

[target.wasm32-unknown-unknown.dependencies]
uuid = { features = ["wasm-bindgen"], git = "https://github.com/zrzka/uuid.git", branch = "feature/wasm" }
```

My `index.js`:

```
const bt = require('balena-temen');

console.log(
    bt.evaluate({
        "id": {
            "$$eval": "uuidv4()"
        }
    })
);
```

[uuidv4() implementation](https://github.com/balena-io-modules/balena-temen/blob/master/src/builtin/function/uuidv4.rs#L9-L11).

Output:

```
{ id: 'cefa2919-ff48-48ef-a231-e13697e23ed2' }
```


Dylan-DPC:
Thanks. But I don't think it is a good idea to depend on docs for messages that it is supported only on certain targets with a feature gate. I'm looking for a better way of doing this.

zrzka:
Did update this PR and gated all v3, v4, v5 modules behind features. If you target `wasm32-...` and you forget to enable either `stdweb` or `wasm-bindgen` feature, it fails to compile:

```
warning: unused manifest key: package.maintenance
   Compiling rand v0.6.1
   Compiling uuid v0.7.1 (https://github.com/zrzka/uuid.git?branch=feature/wasm#0ccfd12f)
   Compiling balena-temen v0.0.20 (/Users/robertvojta/Work/balena/balena-temen)
error[E0599]: no function or associated item named `new_v4` found for type `uuid::Uuid` in the current scope
  --> src/builtin/function/uuidv4.rs:10:22
   |
10 |     Ok(Value::String(Uuid::new_v4().to_hyphenated().to_string()))
   |                      ^^^^^^^^^^^^ function or associated item not found in `uuid::Uuid`
```

src/lib.rs
```diff
@@ -160,11 +160,47 @@ mod std_support;
...
 mod v4;
-#[cfg(feature = "v5")]
+#[cfg(all(
+    feature = "v5",
+    any(
+        not(target_arch = "wasm32"),
+        all(
+            target_arch = "wasm32",
+            any(
+                all(feature = "stdweb", not(feature = "wasm-bindgen")),
```
Dylan-DPC: do we need both features to be mutually exclusive? what will happen if this condition doesn't exist and both features exist?

wasm-pack.log
```diff
@@ -0,0 +1,4 @@
+Nov 27 08:29:43.637 ERRO crate-type must be cdylib to compile to wasm32-unknown-unknown. Add the following to your Cargo.toml file:
```
Dylan-DPC: this file should be added to `.gitignore`

src/lib.rs
zrzka: According to [this](https://github.com/rust-random/rand/blob/master/src/rngs/os.rs#L1004-L1006) and [this](https://github.com/rust-random/rand/blob/master/src/rngs/os.rs#L1107-L1110) I assume that the compilation will fail, because of two `imp` modules. And we should probably add `not(target_os = "emscripten")` as well, to mimick `rand` behavior.

Dylan-DPC: Sounds good to me :+1:

zrzka: Not 100% sure if `not(target_os = "emscripten")` should be added, because [at the top level](https://github.com/rust-random/rand/blob/master/src/lib.rs#L169-L170) only `wasm32` & `stdweb` | `wasm-bindgen` is used.

Dylan-DPC:
Currently this doesn't test the crate on wasm-* but just the features. We need to add another build to the CI so that it will run all the tests (or just the feature-gated ones) on wasm.

src/lib.rs
zrzka: My assumption was wrong.

`cargo build --target wasm32-unknown-unknown --features "stdweb wasm-bindgen"` works and `stdweb` is picked.

Rand contains ...

```
#[cfg(all(target_arch = "wasm32",
          not(target_os = "emscripten"),
feature = "stdweb"))]
```

... which works when `stdweb` is enabled. And it also contains ...

```
#[cfg(all(target_arch = "wasm32",
          not(target_os = "emscripten"),
          not(feature = "stdweb"),
feature = "wasm-bindgen"))]
```

... which utilizes `wasm-bindgen` only if `stdweb` is disabled.

* It means that `stdweb` & `wasm-bindgen` does not need to be mutually exclusive. `rand` can handle this
* Still not sure if `not(target_os = "emscripten")` should be added

Dylan-DPC: I don't think RNGs are supported on emscripten which means if we don't bother about it, then compilation will fail and if we include it in the feature gate, it will compile to nothing.

zrzka:
@Dylan-DPC can you elaborate on tests pls? There're two options:

a) make a crate buildable for `wasm32-...` and that's it, test the features only,
b) make a crate buildable for `wasm32-...` and provide "NPM package" (JS wrapper) and test it with the `wasm-pack test ...`, something like [this](https://github.com/balena-io-modules/balena-temen/blob/master/src/engine/helper.rs#L288)

If a) is the right answer then I don't understand what do you mean with tests. Because you can't do things like `cargo test --target wasm32-unknown-unknown --features "v5 stdweb"` for example.

Also I don't think b) needs to be added.

zrzka:
`rand` is doing [this](https://github.com/rust-random/rand/blob/master/.travis.yml#L127-L163) for WASM, it takes quite a long time and the output is:

```
test result (async): ok. 0 passed; 0 failed
test result (async): ok. 0 passed; 0 failed
```

zrzka:
Did add builds for the `wasm32` arch with features as a minimum.

KodrAus:
Thanks for the PR @zrzka. Personally I'm not a fan of adding transitive `stdweb` or `wasm-bindgen` features to `uuid` for the sake of `rand`. It just passes the problem of figuring out feature gates upstream to the next consumer of `uuid`.

Right now, I think depending on `rand` alongside `uuid` in your project with one of these features set is a better way to go.

If we're going to support random uuids in wasm then I think we should do it without needing any special feature gates.

Dylan-DPC:
> @Dylan-DPC can you elaborate on tests pls? There're two options:
> 
> a) make a crate buildable for `wasm32-...` and that's it, test the features only,
> b) make a crate buildable for `wasm32-...` and provide "NPM package" (JS wrapper) and test it with the `wasm-pack test ...`, something like [this](https://github.com/balena-io-modules/balena-temen/blob/master/src/engine/helper.rs#L288)

This will test only the features, it won't test whether uuid works on the wasm architectures or not, which could lead to false truths.

Dylan-DPC:
@KodrAus I don't think you can do it without the feature gates, and since the rand features are different depending on whether they want stdweb or bindgen. I don't see another way of doing this at this point.

zrzka:
@Dylan-DPC is right :)

Anyway, I have fixed my issue [with this commit](https://github.com/balena-io-modules/balena-temen/pull/41/commits/737b578ae6a0efdbf6a8a197ae5702e5b7ddbbdd), works for me. Ugly as well, but at least it works.

KodrAus:
@zrzka So it looks like this is actually the way we're recommending doing this right now, so if you'd like to re-open this PR as-is then I'm on board with the approach.

Sorry about the misdirection here :) I can live with a solution that has drawbacks if it's consistent with what other libraries are doing.

zrzka:
@KodrAus np, next week.

zrzka:
Restored & reopened as requested. Summary:

* `rand` version bumped
* `stdweb` & `wasm-bindgen` features added, they just transitively enables same features in the `rand` crate
* if anyone wants to use `v3`, `v4`, `v5` on `wasm32` arch, `stdweb` and/or `wasm-bindgen` feature(s) must be enabled otherwise these modules wont be available
* Travis CI builds these combinations for `wasm32-unknown-unknown`
* Note added to `lib.rs` about this behavior

zrzka:
Yep, `stdweb` requires nightly. I reschuffled .travis.yml little bit. Also 1.22 doesn't have support form `wasm32-unknown-unknown`, so, I removed it from the main script.

Dylan-DPC:
Yeah they fixed the issue in [https://github.com/koute/stdweb/commit/f1fc5e320645a9284b3f0e651dba05bcb4720b3b] and it will work on stable once it is released.

Dylan-DPC:
bors: r+

PR #353, opened 2018-12-05
Update md5 0.5 to 0.6

silwol:
# Description
This patch updates the `md5` dependency from 0.5 to 0.6. The API of the `md5` crate is not exposed in `uuid`, just used in one single method which uses it internally, so this is not a semver breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ version = "1"
 
 [dependencies.md5]
 optional = true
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
Thanks. We had #344 open for a while now but i'll merge this PR instead of that.

Dylan-DPC:
bors: r+

PR #355, opened 2019-01-11
fix parser::Error::InvalidCharacter Display message

kinggoesgaming:
**I'm submitting a(n)** bug fix

# Description
Use string instead of `str::Chars`

# Motivation
While working on `uuid_cli` I realized that when an invalid character was provided in a namespace uuid for v3 and v5 variants, the expected characters is shown as a numeric array instead of an array of chars. Not helpful for users. 

# Tests
As its only a error message change, no tests needed

# Related Issue(s)
N/A

### DIFF
--- a/src/parser/core_support.rs
+++ b/src/parser/core_support.rs
@@ -35,8 +35,8 @@ impl fmt::Display for parser::ParseError {
                 index,
             } => write!(
                 f,
-                "expected {:?}, found {} at {}",
-                expected.chars(),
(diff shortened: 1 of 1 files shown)


KodrAus:
The build failure looks like a combination of formatting needed and wasm.

kinggoesgaming:
Ill get to it tonight :+1:

kinggoesgaming:
awaiting on #358

PR #360, opened 2019-01-18
make Debug output for Uuid as hyphenated uuid instead of raw bytes

kinggoesgaming:

# Description
Use `fmt::LowerHex` as the base for `fmt::Debug`

# Motivation
#356 indicates that reading raw byte values for the uuid bytes is not easy for debugging purposes.

# Tests
N/A

# Related Issue(s)
#356


Dylan-DPC:
bors: r+

PR #365, opened 2019-02-11
stabilize const_fn

### LINKED ISSUE #363: Stabilize const_fn feature
`const_fn` feature has been stabilized in Rust. As such we need to adapt our code to support it

kinggoesgaming:
**I'm submitting a(n)** other

# Description
Mark the `const_fn` as a stable feature and remove the compiler feature gate switch.

# Motivation
`const fn $name()` is now stable as of Rust 1.31.0

# Tests
All tests pass

# Related Issue(s)
closes #363

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,5 @@ const_fn = []
 
 # nightly rust
 #------------------------
-# Allow using `const fn`s
 # Nightly marker feature gate
 nightly = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,9 +83,11 @@ wasm-bindgen = ["rand/wasm-bindgen"]
 # since rust 1.26.0
 u128 = ["byteorder"]
 
+# since rust 1.31.0
+const_fn = []
+
 # nightly rust
 #------------------------
 # Allow using `const fn`s
-const_fn = ["nightly"]
 # Nightly marker feature gate
 nightly = []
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -83,9 +83,10 @@ wasm-bindgen = ["rand/wasm-bindgen"]
 # since rust 1.26.0
 u128 = ["byteorder"]
 
+# since rust 1.31.0
+const_fn = []
```
Dylan-DPC: why do we need to keep the feature? since it will be a part of all versions we support and i don't see a reason for any user to opt-in or opt-out of it.

kinggoesgaming: since currently people may be using `uuid` transitively and dont want their code to break. `0.8.0` will remove them (should prolly document that)

KodrAus: Ah that sounds fair. At least before `1.0` we should review and deprecate any features that we don't need anymore. But we have removed unstable features from `lazy-static` in minor versions post `1.0` so shouldn't feel like we have to be stuck with them forever in any case.

kinggoesgaming: well with `0.8.0` it will happen anyhow as we will transition to `2018` edition which will automatically guarantee that `const_fn` feature is available. 1.31.0 introduced `const_fn` and `2018` edition is `1.32.0`

kinggoesgaming:
This should be ready to go

@uuid-rs/uuid

KodrAus:
Failure is unrelated: nightly is broken.

kinggoesgaming:
Thinking on this more, I think I will drop this and instead move to work onto `0.8.0`, which will remove this anyhow.

KodrAus:
Sounds good to me!

PR #367, opened 2019-02-13
make winapi a windows-specific dependency

### LINKED ISSUE #362: uuid now depends on winapi on all platforms
This is unfortunate and we (Fedora) would appreciate if you would mark it as needed only on Windows.

Thanks!

kinggoesgaming:
**I'm submitting a(n)** bug fix

# Description
Using `[target.'cfg()'` format to only link to winapi only on windows

# Motivation
Hey fedora doods, you happy now?

# Tests
tests passing

# Related Issue(s)
closes #362


kinggoesgaming:
hahahaha. will fix this soon

Cargo.toml
```diff
@@ -89,3 +85,7 @@ u128 = ["byteorder"]
 const_fn = ["nightly"]
 # Nightly marker feature gate
 nightly = []
+
+[target.'cfg(all(windows, target_env="msrv"))'.dependencies.winapi]
```
KodrAus: Should this be `msvc`?

KodrAus: Also, do we even need the `target_env` here? If you're building with a GNU toolchain on Windows you might still want `winapi`, right?

kinggoesgaming: > Should this be `msvc`?

oops



> Also, do we even need the `target_env` here? If you're building with a GNU toolchain on Windows you might still want `winapi`, right?

true. will adjust this when I can

kinggoesgaming:
@uuid-rs/uuid this should be good to go when the build checks are done

Dylan-DPC:
bors: r+

kinggoesgaming:
bors r=@Dylan-DPC

ignatenkobrain:
Thanks!

PR #371, opened 2019-02-28
Remove `rand` dependency for v3 and v5

Boddlnagg:
I don't see why the v3 and v5 features depend on the `rand` crate, but maybe I'm missing something?
If not, I would like to get rid of that dependency (which pulls in quite a number of crates transitively).

`cargo test --features "v5 v3"` still passes locally.

I don't think that removing a dependency is a breaking change.


Dylan-DPC:
Thanks. it is fine since we are not using the rand crate in v3 and v5, so this must be some residue that got left behind.

Dylan-DPC:
bors: r+

PR #373, opened 2019-03-03
Add Error type

kinggoesgaming:

# Description
Add `uuid::Error` that sums up `uuid::BytesError` and `parser::ParseError`

# Motivation
Add a way for frameworks to not worry to handing both error types in `uuid` as separate types. Allows for exposing errors that don't explicitly matter until there is a need to handle them.

# Tests

# Related Issue(s)
N/A


src/lib.rs
```diff
@@ -218,6 +218,35 @@ pub struct BytesError {
...
+/// Although specialized error types exist in the crate,
+/// sometimes where particular error type occurred is hidden
+/// until errors need to be handled. This allows to enumerate 
+/// the errors.
+///
+/// [`Uuid`]: struct.Uuid.html
+// TODO: improve the doc
+// BODY: This detail should be fine for initial merge
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum Error {
```
KodrAus: I think we'd be better off with an opaque struct here than an `enum`. Something like:

```rust
pub struct Error(Inner);

enum Inner {
    Bytes(..),
    Parse(..),
}

impl Error {
    pub fn is_bytes(&self) -> bool {
        ..
    }

    pub fn is_parse(&self) -> bool {
        ..
    }
}
```

And making the `BytesError` and `ParseError` types private in our next breaking release in favour of this single error type. I don't think we need a complex error management story for `uuid`.

kinggoesgaming: That does seem like a good idea...

kinggoesgaming: A note on the making the `BytesError` and `parser::ParseError` private: I designed the parse error type in a way so crates like `blog-uuid` and `uuid-b64` can it and have a seamless design. Should I abandon that or should I readjust API accordingly?

kinggoesgaming: @KodrAus

Dylan-DPC: i would prefer readjusting the API first if possible

KodrAus: @kinggoesgaming Ah so that other libraries can try parse a uuid and use the returned error type to maybe build up their own more semantic error to show a user?

kinggoesgaming: @KodrAus that is a good way to put it. :smile:

kinggoesgaming: I will open this PR for review... if we need to alter the API as the way you suggested @KodrAus, it can be done in `0.8.0` as it will affect the error APIs anyways

src/core_support.rs
```diff
@@ -13,13 +13,28 @@ use core::{fmt, str};
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         fmt::LowerHex::fmt(self, f)
     }
 }
 
+impl fmt::Display for super::Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            super::Error::Bytes(ref err) => fmt::Display::fmt(&err, f),
+            super::Error::Parse(ref err) => fmt::Display::fmt(&err, f),
```
Dylan-DPC: since both rhs are the same we can reduce it to using a wildcard or an `|`

src/std_support.rs
```diff
@@ -11,8 +11,17 @@
...
     fn description(&self) -> &str {
         "invalid number of uuid bytes"
     }
 }
+
+impl error::Error for super::Error {
+    fn description(&self) -> &str {
+        match *self {
+            super::Error::Bytes(ref err) => error::Error::description(err),
+            super::Error::Parse(ref err) => error::Error::description(err),
```
Dylan-DPC: same as above

src/core_support.rs
kinggoesgaming: not possible as LHS types in the variants are different

src/std_support.rs
kinggoesgaming: same as above

src/core_support.rs
Dylan-DPC: it may be possible, but fine don't want to drag this PR cause of that.

kinggoesgaming:
dont merge for now

kinggoesgaming:
have to add tests

Dylan-DPC:
bors: r+

PR #380, opened 2019-03-19
fix a tiny mistake in the README generation

kinggoesgaming:
**I'm submitting a(n)** other

# Description
The way the generation will work is that it will use the link according to local Cargo.toml uuid version. So hopefully it will always use the reference to the docs.rs uuid

# Motivation
The Uuid struct wasn't being linked.


Dylan-DPC:
bors: r+

PR #390, opened 2019-03-31
enable adapter tests for encode_{lower, upper}

kinggoesgaming:
**I'm submitting a(n)** other

# Description
`encode_{lower, upper}` tests were not enabled when we enabled the functions.

# Motivation
More tests

# Tests
New tests enabled

# Related Issue(s)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -938,7 +938,7 @@ impl Uuid {
 
 #[cfg(test)]
 mod tests {
-    pub use std;
+    extern crate std;
 
     use super::test_util;
     use crate::prelude::*;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -938,7 +938,7 @@ impl Uuid {
 
 #[cfg(test)]
 mod tests {
-    pub extern crate std;
+    pub use std;
```
KodrAus: Looks like this might be causing the build to fail in no-std environments

kinggoesgaming: nope... tests module always had the declaration.

I think I just have to revert the change and it should be fine

kinggoesgaming: 6cbfb3a fixes this

Dylan-DPC:
bors: r+

PR #389, opened 2019-03-31
const fn

kinggoesgaming:
**I'm submitting a(n)** feature|refactor

# Description
All previous `const fn` functions that had non-`const fn` functions now lost the other ones.

# Motivation
`2018` edition introduced `const fn`s and it better :)

# Tests
Current tests should pass. No new tests added

# Related Issue(s)
#366

PR #394, opened 2019-04-07
Make functions that can be const, const

mehcode:

# Description

Add `const fn` in a few more places than #389. Used `clippy` and its nursery lint to detect them.

# Motivation

More `const fn` the better. 

# Tests

Current tests should pass. No new tests added

# Related Issue(s)


kinggoesgaming:
hmm I thought I missed stuff when I did that PR.. Thanks  @mehcode

PR #395, opened 2019-04-07
Add uuid! macro to produce compile-time Uuids from strings.

mehcode:

# Description

This moves out parsing into its own crate, `uuid-parser`; and, adds a third crate, `uuid-macros`, with a single expression proc-macro `uuid!` that can produce compile-time UUIDs from strings.

As this is a pretty invasive refactor, please let me know if there are changes to the organization you'd like me to make.

# Motivation

Replaces:

```rust
let id = Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000").unwrap();
```

With:

```rust
let id = uuid!("550e8400-e29b-41d4-a716-446655440000");
```

Furthermore, after #394, this enables _const_ UUIDs to be `const` themselves.

```rust
const ID: Uuid = uuid!("550e8400-e29b-41d4-a716-446655440000");
``` 

# Tests
A new test was added to showcase `uuid!` and prove it works.

# Related Issue(s)


kinggoesgaming:
Thanks for this PR. However I am not yet comfortable with including this into the main crate. However we have an inactive crate (uuid-rs/uuid_macro) which was designed for this for this purpose (polish it out and then introduced). I would love if you open this again that

mehcode:
That crate looks similar enough. I hadn't noticed it. Though I don't see how I could improve the crate much (from a quick glance) without the split here to move out `uuid-parser`.

I'd recommend renaming it to `uuid-macros` though. Never know when we might want a second one.

PR #404, opened 2019-05-10
add sentence to `Uuid::to_timestamp` docs specifying format

jonathanstrong:
I'm submitting a small addition to the documentation for `Uuid::to_timestamp`.

# Description

The sentence explains what the timestamp number represents (100ns intervals since 1582-10-15). 

# Motivation

I think it's wise to include since this is not a common timestamp format in other contexts.


kinggoesgaming:
Thanks @jonathanstrong 

bors r+

PR #412, opened 2019-06-19
Implement the default trait.

thedrow:

# Description
The default trait allows us to use Rust's syntax for default values.
It was previously missing.

The default value is `Uuid::nil()`.

# Motivation
I want to allow Rust programmers to use the idioms they are used to with this library.

# Tests

I added a unit test.

# Related Issue(s)
None

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -889,7 +889,7 @@ impl Uuid {
     }
 }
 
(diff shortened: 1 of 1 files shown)


thedrow:
I didn't notice it was implemented elsewhere.
Apologies for the noise.

Issue #406, opened 2019-05-12
Differences in parsing UUIDs from uint128 in Python and Rust


thedrow:
**Describe the bug**
I'm creating a binding of this library to Python for learning purposes and possible production usage.
I've encountered a difference in the values Python provides and the values Rust provides when parsing a UUID from an unsigned 128-bit integer.

**To Reproduce**

When parsing a UUID from an unsigned integer in Python I get the following UUID:
```python
>>> import uuid
>>> uuid.UUID(int=286695482685957555672632549999181678124)
UUID('d7af8a9f-e924-4f96-b36f-9ecea445c62c')
```

However, when I use the following rust code:
```rust
extern crate uuid;
use uuid::Uuid;
let u: uint128 = 286695482685957555672632549999181678124;
let uuid: Uuid = u.into();
println!(uuid.to_hyphenated()
                .encode_lower(&mut Uuid::encode_buffer())
                .to_string());
```

The output is the following UUID: `2cc645a4-ce9e-6fb3-964f-24e99f8aafd7`

**Expected behavior**
I think that the value should be the same in both implementations.
If not, we need to figure out the difference and decide which implementation is correct.

**Screenshots**

**Specifications (please complete the following information):**
- Target: Debug
- Version [e.g. 1.0]: 0.7.4
- Features Enabled: v1, v3, v4, v5, u128

**Additional context**
See the Python implementation [here](https://github.com/python/cpython/blob/master/Lib/uuid.py#L158).

**Other**
N/A

kinggoesgaming:
cc @KodrAus

KodrAus:
Like like endianness? Will dig in properly in a bit, but maybe we’re reversing the bits in that 128bit number so they’re big endian whereas Python isn’t?

kinggoesgaming:
Now I feel silly... thats the first thing I ruled out :P

thedrow:
That's exactly the case.

I did the following:
```rust
            if let Some(i) = int {
                handle = i.swap_bytes().into();
            }
```
and the tests pass. I'm not sure which implementation is correct (or if both are correct).
In any case, this should be documented.

KodrAus:
Yeh, I think because a `uuid` doesn’t have a representation as a 128bit value any semantics we pick will be fairly arbitrary.

I think we should consider deprecating the 128bit conversions.

thedrow:
I disagree.
It's useful at least to me.

KodrAus:
[...]

PR #407, opened 2019-05-26
remove support for u128

kinggoesgaming:
**I'm submitting a(n)** other

# Description
Remove the `u128_support` module and the current implementations for converting from `u128` to `Uuid`s. New alternative will be provided later.

# Motivation
#406 shows that `u128`, number type that is affected by the target's endianness.

# Tests
All tests related to `u128` removed. The rest of the tests should work as expected, as before.

# Related Issue(s)
#406

PR #416, opened 2019-07-04
add endian adapters

kinggoesgaming:

# Description
Add `uuid::adapter::endian::{Big, Little}`.

# Motivation
Currently we have single representation for `uuid::Uuid` in-memory. However people have needed for little endian. Particularly bindings have a requirement of the memory layout (see #406).

# Tests
Currently new tests are needed to be added

# Related Issue(s)
#406

### DIFF
--- a/src/adapter/endian/core_support.rs
+++ b/src/adapter/endian/core_support.rs
@@ -1,7 +1,26 @@
 use byteorder;
 use crate::prelude::*;
 use crate::adapter::endian;
-use rand::AsByteSliceMut;
+
+impl From<u128> for endian::Big {
+    fn from(f: u128) -> Self {
+        let mut bytes: Bytes = [0; 16];
+
+        <byteorder::BE as byteorder::ByteOrder>::write_u128(&mut bytes, f);
+
+        endian::Big(bytes)
+    }
+}
+
+impl From<u128> for endian::Little {
+    fn from(f: u128) -> Self {
+        let mut bytes: Bytes = [0; 16];
+
+        <byteorder::LE as byteorder::ByteOrder>::write_u128(&mut bytes, f);
+
+        endian::Little(bytes)
+    }
+}
 
 impl From<Uuid> for endian::Big {
     fn from(f: Uuid) -> Self {
--- a/src/adapter/endian/core_support.rs
+++ b/src/adapter/endian/core_support.rs
@@ -1,0 +1,22 @@
+use byteorder;
+use crate::prelude::*;
+use crate::adapter::endian;
+use rand::AsByteSliceMut;
+
+impl From<Uuid> for endian::Big {
+    fn from(f: Uuid) -> Self {
+        let bytes = f.0;
+
+        endian::Big(bytes)
+    }
+}
+
+impl From<Uuid> for endian::Little {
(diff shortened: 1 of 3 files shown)


kinggoesgaming:
@uuid-rs/uuid I need your input on this. I want to make sure I am doing this correct before I add serde and other content to this PR

Dylan-DPC:
looks good to me but good to get an opinion from @KodrAus as well

kinggoesgaming:
psss.. @KodrAus opinions?

KodrAus:
Hmm, I think having types specifically for this might be clumsy to use in practice, because what you really want is a `Uuid`, rather than a newtype that wraps it.

What if we have two functions: `from_u128_be` and `from_u128_le` and either deprecate the `From<u128>` impl or note in the docs that it's platform-dependent? (personally I'd be leaning towards deprecating because it's a bit of a footgun)

kinggoesgaming:
> Hmm, I think having types specifically for this might be clumsy to use in practice, because what you really want is a `Uuid`, rather than a newtype that wraps it.
> 
> What if we have two functions: `from_u128_be` and `from_u128_le` and either deprecate the `From<u128>` impl or note in the docs that it's platform-dependent? (personally I'd be leaning towards deprecating because it's a bit of a footgun)

1. `From<u128>` impl was already removed in #407
2. I was thinking that all the `_be` and `_le` methods be moved to respective endian adapters. This includes the `to_fields_{be,le}`.

KodrAus:
What do we expect a caller to do with a `endian::Big` or `endian::Little`? If its only purpose is to act as a point on your way to a `Uuid` then it seems like the difference between:

```rust
use uuid::{Uuid, adapter::endian};

let uuid: Uuid = endian::Big::from(my_u128).into();
```

or

```rust
use uuid::Uuid;

let uuid = Uuid::from_u128_be(my_u128);
```

But if there's something else these adapters would do then maybe they are worth having as separate structs.

kinggoesgaming:
One big one I can think is (de)serialization. Especially given that you that uuid is stored in a particular endian and you don't to risk messing up..

That aside I am fine with providing the methods on `Uuid` itself

kinggoesgaming:
@uuid-rs/uuid can I have a resolution on which way I should go on this so I can finish this PR then finish the few documentation bits I want to have in this for the release?

KodrAus:
I'm happy for you to take this in the direction you feel is most appropriate.

Personally, I think a combination of constructor methods on `Uuid`, and a serialization module like [`adapter::compact`](https://docs.rs/uuid/0.7.4/uuid/adapter/compact/index.html) would be the most consistent with how we already do things.

kinggoesgaming:
> I'm happy for you to take this in the direction you feel is most appropriate.
> 
> Personally, I think a combination of constructor methods on `Uuid`, and a serialization module like [`adapter::compact`](https://docs.rs/uuid/0.7.4/uuid/adapter/compact/index.html) would be the most consistent with how we already do things.

I think that's a good compromise :)

kinggoesgaming:
@KodrAus I should close this... since your PR technically superseded this one

KodrAus:
Ah yes that's right. Thanks @kinggoesgaming, I'll close this one now.

PR #417, opened 2019-07-14
update rand to 0.7

kinggoesgaming:
**I'm submitting a(n)** dependency update

# Description
Updated `rand` `0.6 -> 0.7`

# Motivation
Updates Updates Updates!!!

# Tests
Everything should work as is

# Related Issue(s)
#415

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ version = "0.6"
 
 [dependencies.rand]
 optional = true
-version = "0.6"
+version = "0.7"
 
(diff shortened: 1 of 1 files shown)


kinggoesgaming:
ok so after ACTUALLY reading the error log, we CAN'T update as `getrandom` uses `extern crate std` somewhere in its code and is causing the build to fail instantly. We can leave it for now.

kinggoesgaming:
`rand@0.7` needs `msrv 1.32.0` :(

PR #424, opened 2019-08-22
:arrow_up: rand

matklad:
**I'm submitting a(n)** other

# Description

Update rand to the next major version (0.7)

# Motivation

avoid rand duplication across ecosystem

# Tests
`cargo test` locally works

# Related Issue(s)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
     - cargo build --target wasm32-unknown-unknown --features "v3 wasm-bindgen"
     - cargo build --target wasm32-unknown-unknown --features "v4 wasm-bindgen"
     - cargo build --target wasm32-unknown-unknown --features "v5 wasm-bindgen"
-  - rust: 1.31.0
+  - rust: 1.32.0
     script:
     - cargo test --features "serde std v4"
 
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
Thanks do look at https://github.com/uuid-rs/uuid/pull/417

matklad:
Yeah, rand bumping msrv to 1.32.0 is unfortunate. That said, 1.32.0 is just one version ahead of 1.31.0, and rand is pretty widely used anyway

Dylan-DPC:
I was referring to [this](https://github.com/uuid-rs/uuid/pull/417#issuecomment-511174405). Not sure if it still holds true. We are blocked on https://github.com/rust-random/getrandom/issues/4

matklad:
Hm, I am confused... How it worked with rand 0.6? Surely, to generate a random number for UUID V4 one needs to call into OS for entropy?

matklad:
So, I've looked more into this, and it seems like we are all set?

rand indeed needs std to make `thread_local` work, and UUID already depends on rand with default features (ie, std is always enabled for rand, with old 0.6 version or with new 0.7). 

I might be missing something obvious, but std/no_std seems completely irrelevant.

The only problem with rand is that its MSRV is 1.32, and I argue it's fine to bump MSRV for UUID as well, blaming rand :D

Dylan-DPC:
Thanks. Sorry for the confusion. Never got time to discuss what exactly the no std problem was 😆

PR #427, opened 2019-10-02
Breaking API features and refactoring for `0.8`

KodrAus:
**I'm submitting a(n)** (|feature|refactor|)

# Description

## Modules and Errors

This PR is a broad refactoring of our error types and module layout. The general pattern I've gone for is:

- Define specific error types in submodules. These are currently private.
- Collect these specific error types in an opaque root `Error`. All methods return this single `Error` type so it's easier for consumers to carry `uuid::Error`s in their own code.

It'll also include some implementations for open PRs as we've been discussing. I imagine we'll want to spend time working through these changes :)

I've hidden our `prelude` module for now, because I'm not sure it's something we'll want to stabilize with (it's only got a few bits in it afterall).

## 128bit support

Re-enables support for 128-bit numbers in the form of constructors on `Uuid`, following the pattern that method names without an endian suffix are BE.

## No-std

Refactors our `no-std` support so we always mark the crate as `no-std` so our std imports are always the same. This simplifies our std/core support so we need fewer modules.

## Timestamps

Includes the design proposed in #405 for timestamp support originally implemented by @jonathanstrong.

# Related Issue(s)

- #406

# Related PR(s)

- #405 
- #416


KodrAus:
Since this is a pretty big PR I'm trying to keep the individual commits clean so it should be easier to follow along.

KodrAus:
This TODO bot is a bit of a nuisance...

kinggoesgaming:
(about the todo bot, I think we should keep **until** 1.0, at which point we just remove it)

kinggoesgaming:
Also @KodrAus this is still a WIP right?

KodrAus:
@kinggoesgaming I think this is ready for a review now :+1:

There’s definitely more docs work to do and we can consider how to expose richer info out of our errors but I don’t think that needs to block a review or possible merge.

PR #439, opened 2019-10-21
Introduce Clippy Github Actions

kinggoesgaming:
**I'm submitting a(n)** other

# Description
Moves clippy tests over to Github Actions

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -54,13 +54,3 @@ jobs:
        with:
          command: clippy
          args: --features "guid serde slog std v1 v3 v4 v5"
-  build:
-
-    runs-on: ubuntu-latest
-
-    steps:
-    - uses: actions/checkout@v1
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: cargo test --verbose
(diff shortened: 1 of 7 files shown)


kinggoesgaming:
You will notice some `#[allow(..)]`s. we need to address them at some point and figure out if they need to exist or not

KodrAus:
Do you think we need to run clippy on multiple platforms?

kinggoesgaming:
I think in particular we need one for Windows as guid feature is a thing

I can remove macOS build, but quite honestly currently the build times are not too bad

kinggoesgaming:
@KodrAus / @DylanDPC this is ready to go

.github/workflows/clippy.yml
```diff
@@ -0,0 +1,41 @@
...
+       with:
+         components: clippy
+         override: true
+         profile: minimal
+         toolchain: 1.32.0
+     - name: Run `cargo clippy`
+       uses: actions-rs/cargo@v1
+       with:
+         command: clippy
+         args: --features "guid serde slog std v1 v3 v4 v5" -- -D warnings
```
Dylan-DPC: do we need all the features activated for windows builds as well?

kinggoesgaming: Tbh it doesn’t affect the overall build time much... we can see later if we need to switch it around

Dylan-DPC:
bors: r+

PR #447, opened 2020-01-17
Use getrandom crate for uuid

ammgws:
**I'm submitting a(n)** other

# Description

Use `getrandom` crate directly instead of `rand` for generating UUIDs using v4.

# Motivation

Reduce the number of dependencies used by the crate.

Running the example in the README: 
- Before (10): c2-chacha, cfg-if, getrandom, libc, ppv-lite86, rand, rand_chacha, rand_core, rand_hc, wasi  
- After (4): cfg-if, getrandom, libc, wasi

# Tests

`cargo test` locally and test usage in a simple program.

# Related Issue(s)

# Other

Wasn't sure whether I should bump the version here or not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,11 +54,11 @@ status = "actively-developed"
 [badges.travis-ci]
 repository = "uuid-rs/uuid"
 
-[dependencies.md5]
+[dependencies.getrandom]
 optional = true
-version = "0.7"
+version = "0.1"
 
-[dependencies.rand]
+[dependencies.md5]
 optional = true
 version = "0.7"
 
@@ -92,12 +92,12 @@ version = "1.0.56"
 default = ["std"]
 guid = ["winapi"]
 std = []
-stdweb = [ "rand/stdweb" ]
+stdweb = ["getrandom"]
 v1 = []
 v3 = ["md5"]
-v4 = ["rand"]
+v4 = ["getrandom"]
 v5 = ["sha1"]
-wasm-bindgen = ["rand/wasm-bindgen"]
+wasm-bindgen = ["getrandom"]
 
 [target.'cfg(windows)'.dependencies.winapi]
 optional = true
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ To create a new random (V4) UUID and print it out in hexadecimal form:
 
 use uuid::Uuid;
 
-fn main() {
+fn main() -> Result<(), Box<std::error::Error> {
     let my_uuid = Uuid::new_v4().unwrap();
     println!("{}", my_uuid);
 }
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ To create a new random (V4) UUID and print it out in hexadecimal form:
 use uuid::Uuid;
 
 fn main() {
-    let my_uuid = Uuid::new_v4();
+    let my_uuid = Uuid::new_v4().unwrap();
     println!("{}", my_uuid);
 }
 ```
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ To create a new random (V4) UUID and print it out in hexadecimal form:
 use uuid::Uuid;
 
 fn main() -> Result<(), Box<std::error::Error> {
-    let my_uuid = Uuid::new_v4().unwrap();
+    let my_uuid = Uuid::new_v4()?;
     println!("{}", my_uuid);
 }
 ```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,7 +94,7 @@
 //! use uuid::Uuid;
 //!
 //! fn main() {
-//!     let my_uuid = Uuid::new_v4();
+//!     let my_uuid = Uuid::new_v4().unwrap();
 //!     println!("{}", my_uuid);
 //! }
 //! ```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,8 +93,8 @@
 //!
 //! use uuid::Uuid;
 //!
-//! fn main() {
-//!     let my_uuid = Uuid::new_v4().unwrap();
+//! fn main() -> Result<(), Box<std::error::Error> {
+//!     let my_uuid = Uuid::new_v4()?;
 //!     println!("{}", my_uuid);
 //! }
 //! ```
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -30,7 +30,7 @@ homepage = "https://github.com/uuid-rs/uuid"
 name = "uuid"
 readme = "README.md"
 repository = "https://github.com/uuid-rs/uuid"
-version = "0.8.1" # remember to update html_root_url in lib.rs
+version = "0.8.2" # remember to update html_root_url in lib.rs
```
kinggoesgaming: Leave this unchanged. We use a separate PR to handle the version update just before release

Cargo.toml
```diff
@@ -54,6 +54,10 @@ status = "actively-developed"
 [badges.travis-ci]
 repository = "uuid-rs/uuid"
 
+[dependencies.getrandom]
```
kinggoesgaming: I see you are adding the dependency but you didn't remove `rand`. Is this intentional or just a miss?

src/lib.rs
```diff
@@ -130,7 +130,7 @@
 #![doc(
     html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
     html_favicon_url = "https://www.rust-lang.org/favicon.ico",
-    html_root_url = "https://docs.rs/uuid/0.8.1"
+    html_root_url = "https://docs.rs/uuid/0.8.2"
```
kinggoesgaming: Same here. We handle it before release

src/v4.rs
```diff
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use rand;
+use getrandom::getrandom;
```
kinggoesgaming: `uuid` uses the import module only policy. As `getrandom` is a crate and we are using `2018` edition, we don't need this line at all now.

```suggestion
```

src/v4.rs
```diff
@@ -23,12 +23,8 @@ impl Uuid {
     ///
     /// [`rand`]: https://crates.io/crates/rand
     pub fn new_v4() -> Self {
-        use rand::RngCore;
-
-        let mut rng = rand::thread_rng();
         let mut bytes = [0; 16];
-
-        rng.fill_bytes(&mut bytes);
+        getrandom(&mut bytes).expect("RNG failure!");
```
kinggoesgaming:
```suggestion
        getrandom::getrandom(&mut bytes).expect("RNG failure!");
```

kinggoesgaming: Another issue: you are using `.expect(...)`. I assume it is to maintain the signature of the constructor. But since a `1.0` would be a major release anyways, we can break it. You may have to adjust the error itself to support this though.

Cargo.toml
ammgws: Reverted. Thanks.

src/lib.rs
ammgws: Reverted. Thanks.

Cargo.toml
ammgws: It's still needed for the `stdweb` and `wasm-bindgen` features.

src/v4.rs
ammgws: Done.

src/v4.rs
ammgws: Done.

ammgws: I had a go at it. Could you please see if the changes I made are along the lines of what you were thinking?

src/v4.rs
```diff
@@ -21,19 +20,17 @@ impl Uuid {
     /// let uuid = Uuid::new_v4();
     /// ```
     ///
+    /// [`getrandom`]: https://crates.io/crates/getrandom
     /// [`rand`]: https://crates.io/crates/rand
-    pub fn new_v4() -> Self {
-        use rand::RngCore;
+    pub fn new_v4() -> Result<Uuid, getrandom::Error> {
```
kinggoesgaming:
```suggestion
    // TODO: change signature to support uuid's Error.
    pub fn new_v4() -> Result<Uuid, getrandom::Error> {
```

src/v4.rs
```diff
@@ -1,12 +1,11 @@
 use crate::prelude::*;
-use rand;
 
 impl Uuid {
     /// Creates a random UUID.
     ///
-    /// This uses the [`rand`] crate's default task RNG as the source of random
-    /// numbers. If you'd like to use a custom generator, don't use this
-    /// method: use the `rand::Rand trait`'s `rand()` method instead.
+    /// This uses the [`getrandom`] crate to utilise the operating system's RNG
```
kinggoesgaming: @KodrAus should we expose the fact that we are using the `getrandom` crate implementation in the docs?

src/lib.rs
```diff
@@ -94,7 +94,7 @@
 //! use uuid::Uuid;
 //!
 //! fn main() {
```
kinggoesgaming:
```suggestion
//! fn main() -> Result<(), Box<std::error::Error> {
```

src/lib.rs
```diff
@@ -94,7 +94,7 @@
...
-//!     let my_uuid = Uuid::new_v4();
+//!     let my_uuid = Uuid::new_v4().unwrap();
```
kinggoesgaming:
```suggestion
//!     let my_uuid = Uuid::new_v4()?;
```

README.md
```diff
@@ -96,7 +96,7 @@ To create a new random (V4) UUID and print it out in hexadecimal form:
 use uuid::Uuid;
 
 fn main() {
```
kinggoesgaming:
```suggestion
fn main() -> Result<(), Box<std::error::Error> {
```

README.md
```diff
@@ -96,7 +96,7 @@ To create a new random (V4) UUID and print it out in hexadecimal form:
...
-    let my_uuid = Uuid::new_v4();
+    let my_uuid = Uuid::new_v4().unwrap();
```
kinggoesgaming:
```suggestion
    let my_uuid = Uuid::new_v4()?;
```

src/v4.rs
```diff
@@ -1,12 +1,11 @@
...
+    /// as the source of randomnumbers. If you'd like to use a custom generator,
```
kinggoesgaming:
```suggestion
    /// as the source of random numbers. If you'd like to use a custom generator,
```

kinggoesgaming:
Thanks @ammgws for working on this :)

ammgws:
Shall I rebase this?

PR #449, opened 2020-02-06
make `Uuid::from_guid` infallible

### LINKED ISSUE #446: from_guid should be infallible
**Describe the bug**
`Uuid::from_guid` return type is [declared](https://github.com/uuid-rs/uuid/blob/257b0e57faf6fa3eb248052055c66c703c6d9cd8/src/winapi_support.rs#L9) as `Result<Uuid, Error>`. But this function returns errors only if `Uuid::from_fields_le()` failed. And that function only returns error [if](https://github.com/uuid-rs/uuid/blob/257b0e57faf6fa3eb248052055c66c703c6d9cd8/src/builder/mod.rs#L127) `d4` is not 8 bytes long. However in `from_guid` it is always 8 bytes long. In fact this function is always successful.  

**Expected behavior**
This function should return `Uuid`.

**Specifications**
- Version: 0.8.1, git master

kinggoesgaming:
closes #446

### DIFF
--- a/src/winapi_support.rs
+++ b/src/winapi_support.rs
@@ -3,18 +3,17 @@ use winapi::shared::guiddef;
 
 #[cfg(feature = "guid")]
 impl Uuid {
-    /// Attempts to create a [`Uuid`] from a little endian winapi `GUID`
+    /// Converts a little endian winapi `GUID` into a [`Uuid`]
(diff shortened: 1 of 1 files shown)


Dylan-DPC:
bors: r+

PR #467, opened 2020-03-19
Fix example in README.md

heckad:
Fix


kinggoesgaming:
We are currently experiencing problems with Travis 

I’ll work tomorrow on moving GHActions and then work on merging this

kinggoesgaming:

Granted this is an easy thing, that technically should not involve testing (this test is covered in rustdocs, in the source) but I want to keep it on structure.

kinggoesgaming (approved):
Thank you

PR #468, opened 2020-03-23
make uuid structs repr(transparent)

### LINKED ISSUE #463: make Uuid #[repr(transparent)]
Based on #461 by @DianaNites

---
The crate supports `no_std`, but the Uuid type isn't FFI safe/`repr(transparent)`? There doesn't seem to be any reason for it not to be. That attribute is stable since `1.28.0`, which is below the crates MSRV.

kinggoesgaming:

# Description
Make the following structs `#[repr(transparent)]`:
* `Uuid`
* `Hyphenated`
* `HyphenatedRef`
* `Simple`
* `SimpleRef`
* `Urn`
* `UrnRef`

# Tests
Tests should pass as is

# Related Issue(s)
closes #463


kinggoesgaming:

But I am definitely want this merged ASAP.

kinggoesgaming:
@uuid-rs/uuid  :)

KodrAus:
I'm on board with marking `Uuid` as `#[repr(transparent)]`. I'm not sure we need to commit to marking its display adapters as transparent though, do we? I could imagine wanting to capture some other formatting information in some of them one day, like what case to use for `a` - `f`.

kinggoesgaming:
> 
> 
> I'm on board with marking `Uuid` as `#[repr(transparent)]`. I'm not sure we need to commit to marking its display adapters as transparent though, do we? I could imagine wanting to capture some other formatting information in some of them one day, like what case to use for `a` - `f`.

@KodrAus fair enough! I will remove the attribute from the adapters

kinggoesgaming:
@KodrAus anything else?

KodrAus (approved):
This looks good to me!

PR #477, opened 2020-07-05
Add wasm32-wasi support for V3, V4, and V5

sunfishcode:

# Description

When the wasm32 architecture is used with WASI, md5, getrandom, and sha1
are all available, so V3, V4, and V5 can be enabled.

# Motivation

I'm porting an application that uses `uuid` to WASI and would like to add support.

# Tests

`cargo test --target=wasm32-wasi --all-features` passes.


kinggoesgaming:

I am not sure how we handle post the shift to Github Actions complete. (Travis is completely broken for us now so...)

PR #489, opened 2020-08-31
Clarify documentation around the relation to the rand crate

carols10cents:
**I'm submitting a(n)** other - documentation improvement

# Description

The rand crate changed its trait names, so these docs were referencing a
trait that doesn't exist in the current version of rand. Change to be a
bit more vague to avoid this getting out of date again.

Also "use rand instead" wasn't very clear because "rand" doesn't contain
something clearly equivalent like `get_uuid_v4`. To be clearer, add a
reference to the `Builder::from_bytes` function in this crate.

# Motivation

It took me a few minutes to figure out what this documentation was saying, and I'd like to save someone else those few minutes in the future.

# Tests

N/A

# Related Issue(s)

None

### DIFF
--- a/src/v4.rs
+++ b/src/v4.rs
@@ -5,7 +5,9 @@ impl Uuid {
     ///
     /// This uses the [`getrandom`] crate to utilise the operating system's RNG
     /// as the source of random numbers. If you'd like to use a custom generator,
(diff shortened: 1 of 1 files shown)


kinggoesgaming:
Hi can you merge changes from master and then I can merge

carols10cents:
Done!

PR #495, opened 2020-09-11
cleanup: use better-maintained sha-1 crate instead of sha1 crate

durin42:
This removes a "duplicate" dependency in the tree of pyoxidizer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,9 +67,9 @@ default-features = false
 optional = true
 version = "1.0.56"
 
-[dependencies.sha1]
+[dependencies.sha-1]
 optional = true
-version = "0.6"
+version = "0.9.1"
 
 [dependencies.slog]
 optional = true
@@ -95,10 +95,9 @@ stdweb = ["getrandom", "getrandom/js"]
 v1 = []
 v3 = ["md5"]
 v4 = ["getrandom"]
-v5 = ["sha1"]
+v5 = ["sha-1"]
 wasm-bindgen = ["getrandom", "getrandom/js"]
 
 [target.'cfg(windows)'.dependencies.winapi]
 optional = true
 version = "0.3"
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,9 +58,9 @@ repository = "uuid-rs/uuid"
 optional = true
 version = "0.2.0"
 
-[dependencies.md5]
+[dependencies.md-5]
 optional = true
-version = "0.7"
+version = "0.9.1"
 
 [dependencies.serde]
 default-features = false
@@ -93,7 +93,7 @@ guid = ["winapi"]
 std = []
 stdweb = ["getrandom", "getrandom/js"]
 v1 = []
-v3 = ["md5"]
+v3 = ["md-5"]
 v4 = ["getrandom"]
 v5 = ["sha-1"]
 wasm-bindgen = ["getrandom", "getrandom/js"]
--- a/src/v3.rs
+++ b/src/v3.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use md5;
+use md5::Digest;
 
 impl Uuid {
     /// Creates a UUID using a name from a namespace, based on the MD5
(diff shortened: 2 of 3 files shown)


alex:
sha1 has a higher MSRV than UUID currently does.

newpavlov:
`sha-1 v0.8` has MSRV [equal to 1.21](https://github.com/RustCrypto/hashes/tree/sha1-v0.8.2#minimum-supported-rust-version-msrv). Ideally together with such change it's worth to replace `md5` with `md-5` as well, though maintainers have previously [rejected](https://github.com/uuid-rs/uuid/pull/93) my PR with such changes. :/

KodrAus:
It looks like the original concern came from the weight of those dependencies? I had a look through them and they didn't seem too weighty to me.

I'd be in favour of also bundling `md-5` into this PR if you're happy to do that.

durin42:
It was mostly I noticed the two implementations of SHA1 as I was importing crates into a monorepo and figured I should try and consolidate down onto the maintained-looking one. If you're raising your MSRV I'm happy to handle both sha1 and md5, but if you want to keep it low please feel encouraged to reject this PR.

Thanks!

KodrAus:
We've actually already raised it, so please feel free to roll in both!

KodrAus:
@durin42 I just pushed a quick merge commit to your branch. Hope you don't mind!

durin42:
Not at all! I'll try and get the md5 upgrade done too over the weekend. Right now my keyboard time is somewhat limited. :)

durin42:
I've rebased and added a commit for upgrading md5 to md-5, but I got stuck figuring out how to convert the GenericArray into a [u8;16] - please let me know if there's a trick I'm missing!

durin42:
I'm not sure what's up with the `cargo clippy` presubmit check. Works here?

Cargo.toml
```diff
@@ -58,18 +58,18 @@ repository = "uuid-rs/uuid"
 optional = true
 version = "0.2.0"
 
-[dependencies.md5]
+[dependencies.md-5]
 optional = true
-version = "0.7"
+version = "0.9.1"
```
newpavlov, line 63: It could be worth to start with v0.8 versions of `md-5` and `sha-1` to avoid bumping MSRV. Updating versions can be done later in a separate PR.

src/v3.rs
```diff
@@ -20,15 +20,14 @@ impl Uuid {
...
-        context.consume(name);
+        hasher.update(namespace.as_bytes());
+        hasher.update(name);
 
-        let computed = context.compute();
-        let bytes = computed.into();
+        let computed = hasher.finalize();
 
-        let mut builder = crate::Builder::from_bytes(bytes);
+        let mut builder = crate::Builder::from_slice(&computed[..16]).unwrap();
```
newpavlov, line 30: IIUC you don't need the slicing here, i.e. you can replace `&computed[..16]` with `&computed`. Same goes for SHA-1.

src/v3.rs
```diff
@@ -20,15 +20,14 @@ impl Uuid {
     /// [`NAMESPACE_URL`]: #associatedconstant.NAMESPACE_URL
     /// [`NAMESPACE_X500`]: #associatedconstant.NAMESPACE_X500
     pub fn new_v3(namespace: &Uuid, name: &[u8]) -> Uuid {
-        let mut context = md5::Context::new();
+        let mut hasher = md5::Md5::new();
```
newpavlov, line 23: Use `use md5::{Md5, Digest};` impoer to avoid duplication of `md5` here? Same goes for SHA-1.

Cargo.toml
KodrAus: @newpavlov We've actually already bumped the MSRV so there's no reason not to track the latest versions here.

KodrAus:
It looks like it's trying to read configuration files for dependencies and is failing on one of them:

```
error reading Clippy's configuration file `C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\typenum-1.12.0\clippy.toml`
```

We can probably just remove the `clippy` check in CI if you want to remove the [`clippy.yaml`](https://github.com/uuid-rs/uuid/blob/master/.github/workflows/clippy.yml) file.

KodrAus:
I'm not familiar with `generic_array` but took a quick look and it looks like there's a `From` impl from `GenericArray` into ` [T; N]` if we bump our MSRV to `1.41.0`, which I'm personally ok with. `1.41.0` was released almost 12 months ago.

KodrAus:
Hi @durin42 :wave:

I've ended up doing this in #526. Thanks for working on this!

Issue #472, opened 2020-05-24
Add impl AsRef<[u8]> for Uuid


lassade:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Add some along this lines:
```
impl AsRef<[u8]> for Uuid {
    #[inline]
    fn as_ref(&self) -> &[u8] {
        &self.as_bytes()[..]
    }
}
```

**Is it blocking?**
No

**Describe alternatives you've considered**
I tried:
```
trait AsBytes { /*...*/ }
impl<T: AsRef<[u8]>> AsBytes for T { /*...*/ }
impl AsBytes for uuid::Uuid { /*...*/ }
```
but got:
```
error[E0119]: conflicting implementations of trait `AsBytes` for type `uuid::Uuid`: ...
note: upstream crates may add a new impl of trait `std::convert::AsRef<[u8]>` for type `uuid::Uuid` in future versions
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a5cd3e47bec30ab4c71ad4270dfb3f04

**Additional context**
It feels this should be implemented for to cover the Interoperability princible.  #191

Waridley:
This error is indicating that UUID **does** implement `AsRef<[u8]>`, thus your `impl AsBytes for uuid::Uuid { /*...*/ }` is redundant because it's already covered by the impl for `T: AsRef<[u8]>`. 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=41676e06eebccf5753fb17190310db5f

lassade:
You are mistaken the line `println!("{:?}", uuid::Uuid::new_v4().as_bytes());` doesn't call the `as_bytes  from the `AsBytes` trait`, instead it refers to this method [as_bytes](https://docs.rs/uuid/0.8.1/uuid/struct.Uuid.html#method.as_bytes) defined by some Uuid impl;

Changing it's name to `as_bytes1` show it
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2cbe9a66668ac2f016eac6326e5f4e75

Waridley:
Ah, whoops.

KodrAus:
Adding an `AsRef<[u8]> for Uuid` seems reasonable to me. Would somebody like to PR it?

PR #498, opened 2020-10-30
[WIP] Add support for zerocopy traits via a feature.

mwanner:
I'm submitting a feature proposal.

# Description

Allows `uuid::Uuid` to be read from or serialized to a byte slice without copying by using the zerocopy crate.

# Motivation

I like the abstraction that zerocopy provides and UUIDs are often embedded in other structs I'd like to use zerocopy's traits on.  Hiding this behind a feature that needs to be explicitly enabled should keep the general cost of this very low for the vast majority of users of uuid that do not use zerocopy.

# Tests

Tests missing, therefore marked WIP.

# Related Issue(s)

Not aware of any, possibly somewhat related to #472, though.  Clearly adding another variant of `as_bytes()` via the `zerocopy::AsBytes` trait.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ repository = "https://github.com/uuid-rs/uuid"
 version = "0.8.1" # remember to update html_root_url in lib.rs
 
 [package.metadata.docs.rs]
-features = [ "guid", "serde", "slog", "v1", "v3", "v4", "v5" ]
+features = [ "guid", "serde", "slog", "v1", "v3", "v4", "v5", "zerocopy" ]
 default-target = "x86_64-pc-windows-msvc"
 
 [package.metadata.playground]
@@ -74,6 +74,11 @@ version = "0.6"
 [dependencies.slog]
 optional = true
 version = "2"
+
+[dependencies.zerocopy]
+default-features = false
+optional = true
+version = "0.3.0"
 
 [dev-dependencies.bincode]
 version = "1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,6 +197,8 @@ mod v4;
 mod v5;
 #[cfg(all(windows, feature = "winapi"))]
 mod winapi_support;
+#[cfg(feature = "zerocopy")]
+use zerocopy::{AsBytes, FromBytes, Unaligned};
 
 use crate::std::{fmt, str};
 
@@ -259,7 +261,15 @@ pub enum Variant {
 }
 
 /// A Universally Unique Identifier (UUID).
+#[cfg(not(feature = "zerocopy"))]
 #[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub struct Uuid(Bytes);
+
+/// A Universally Unique Identifier (UUID).
+#[cfg(feature = "zerocopy")]
+#[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd,
+    AsBytes, FromBytes, Unaligned)]
 #[repr(transparent)]
 pub struct Uuid(Bytes);
 


KodrAus:
Thanks for the PR @mwanner! Since we’re looking to stabilize this crate I don’t think we can introduce any unstable public dependencies here. This seems like something that would be covered by the [safer transmute](https://github.com/rust-lang/rfcs/pull/2981) API, but that’s still probably a ways off.

Do you know if `zerocopy` has any attributes you can apply to fields of types deriving one of its attributes to tweak how they’re implemented?

mwanner:
> Thanks for the PR @mwanner! Since we’re looking to stabilize this crate I don’t think we can introduce any unstable public dependencies here.

Would implementing the trait for `Uuid` as part of zerocopy itself be a better way forward?  (Not entirely sure that's possible at all, but...)

> This seems like something that would be covered by the [safer transmute](https://github.com/rust-lang/rfcs/pull/2981) API, but that’s still probably a ways off.

Thanks for this link, I wasn't aware of this effort.

> Do you know if `zerocopy` has any attributes you can apply to fields of types deriving one of its attributes to tweak how they’re implemented?

Not sure if that's what you're asking for, but there's e.g. `U32<LittleEndian>` providing means to read from and write to memory with necessary byte swapping.  Clearly nothing fancier as would be required for different versions of UUIDs.  However, it seems to me `Uuid` is internally represented as `[u8; 16]` and a `const Uuid` could in theory be created from a `&[u8; 16]` w/o any copying.

mwanner:
> Since we’re looking to stabilize this crate I don’t think we can introduce any unstable public dependencies here.

Well, it's an optional dependency.  Should this trait better be provided by `zerocopy` itself, in your opinion?  (Not entirely sure that's possible.)

> This seems like something that would be covered by the [safer transmute](https://github.com/rust-lang/rfcs/pull/2981) API, but that’s still probably a ways off.

Thanks for this link, I wasn't aware of these efforts.

> Do you know if `zerocopy` has any attributes you can apply to fields of types deriving one of its attributes to tweak how they’re implemented?

There's the `byteorder` based stuff like e.g. `U32<LittleEndian>` allowing to transparently do the required byte swapping while reading from or writing to memory.  However, nothing that would allow handling different versions of UUIDs.

KodrAus:
The `Uuid::from_bytes` and `Uuid::as_bytes` methods are both `const`, and on the `master` branch we do mark `Uuid` as `#[repr(transparent)]` so you should be able to safely transmute it from a `[u8; 16]`, `u128` etc, but only safely back into a `[u8; 16]`.

I don't think we could create a const `Uuid` from a `&[u8; 16]` without copying, but we should be able to create a `&Uuid` from a `&[u8; 16]`. Is that what you were meaning?

mwanner:
> The `Uuid::from_bytes` and `Uuid::as_bytes` methods are both `const`, and on the `master` branch we do mark `Uuid` as `#[repr(transparent)]` so you should be able to safely transmute it from a `[u8; 16]`, `u128` etc, but only safely back into a `[u8; 16]`.
> 
> I don't think we could create a const `Uuid` from a `&[u8; 16]` without copying, but we should be able to create a `&Uuid` from a `&[u8; 16]`. Is that what you were meaning?

Yes, sorry, that's what I meant.  Is there currently a safe variant of the following thanks to `repr(transparent)`?

```
pub fn from_u8_array_ref(data: &[u8; 16]) -> &Uuid {
    unsafe { transmute(data) }
}
```

KodrAus:
It looks like we don't currently have one, but we could add something like this in `uuid`:

```rust
impl Uuid {
    pub fn ref_from_bytes(data: &[u8; 16]) -> &Uuid;
}
```

and deal with the unsafety in there. I'm not sure whether it would be valid to cast other types just yet, but it could be possible to also do `ref_from_u128` since `Uuid`s alignment is looser than `u128`s?

sivadeilra:
Hi, folks.  I'd like to help drive this PR forward.  I didn't think to check the PR list for `uuid`, and I ended up writing almost exactly the same PR as this one!  :)

@KodrAus, could you help me understand this?  "Since we’re looking to stabilize this crate I don’t think we can introduce any unstable public dependencies here."  Is the requirement here that all dependencies of `uuid` also be stable, even if those dependencies are optional features?

Also, can you clarify what you mean by "stabilize"?  My impression of the `zerocopy` crate is that it has essentially stabilized, but do you mean a more formal process involving the Rust Project itself?

Supporting `zerocopy` is really important for my team -- I am the dev lead for a team at Microsoft, working on Rust support in Windows.  And we use a lot of GUIDs / UUIDs, of course.  :)  The workarounds listed above in this conversation don't really work for us, because we need to be able to embed UUIDs into other data structures, and those data structures all use `zerocopy` traits to declare that the entire structure can be safely transmuted.

KodrAus:
Hey @sivadeilra :wave:

The constraint we're working with is that any dependency that affects `uuid`'s public API (so types in public functions, and trait implementations on public types) need to have a stable version (that is, `1.0+`) otherwise we can't bump them without a corresponding major version bump to `uuid`, or a commitment to implementing all minor versions that come up before `1.0`. I wouldn't want to appear to promise `zero-copy` support by including it now while `uuid` is unstable and then having to back it out later :)

We have actually since added a `#[repr(transparent)]` attribute to `Uuid` though so the best path forward for your code might be to create a newtype wrapper over `Uuid` that implements the necessary traits. I know it's not as nice as just having everything line up.

sivadeilra:
Hi, thanks for responding.

> We have actually since added a `#[repr(transparent)]` attribute [...]

Unfortunately, a newtype over `Uuid` won't work, because the `zerocopy` traits require that the inner type also implement the same traits, similar to how `Send` requires the same. And like `Send`, these traits are "unsafe" in the sense that they cannot be implemented in verifiable code, only derived using `#[derive]`.

I've worked with the owner of `zerocopy` before on a few changes. My perception is that the crate is already considered "stable", and they already guarantee API stability, but obviously it's using pre-1.0 version numbers. If I contact the crate owner, and he agrees to raise the version number to 1.0 (and to meet the backward-compatibility stability guarantees), would that meet the requirements for adding this to `Uuid`?

sivadeilra:
Ping, on my previous comment.

sivadeilra:
I think the `zerocopy-unstable` option probably makes the most sense. It seems cleaner to me to put the implementations with the type, rather than in the crates that define the traits.

A bunch of us have been trying to push forward on supporting safe transmute traits directly in `core`, but there really hasn't been much momentum on that lately. I think traits like `AsBytes` really should be in `core`, since they define fundamental aspects of traits, and should eventually be auto-derived in the same way that `Send` and `Sync` are.

Would you accept the option of defining a feature for `zerocopy-unstable`, in the near term?  If we ever get the safe-transmute story straightened out, then `uuid` could move to that, and we could remove the `zerocopy-unstable` feature or just make it a no-op.

KodrAus:
@sivadeilra I think I’d prefer we define the impl in `zerocopy` once we’ve stabilized `uuid` if that’s at all possible. At least until `zerocopy` becomes stable then we could move it here. We can fall back to a `zerocopy-unstable` feature here if that’s not possible, but should probably also combine it with a direct `RUSTFLAGS` cfg so you have to opt-in to potential breakage more strongly.

It’s a bit of a shame just how disruptive semver bumps can be, even if there’s very little difference, but that’s the nature of it 🙂

KodrAus:
I've opened a PR in #538 that adds unstable `zerocopy` support here. In order for it to work you need to add a `RUSTFLAGS` cfg, which makes it difficult for libraries to use, but applications can. If that's not suitable for your needs then I think adding support in `zerocopy` directly will be the way to go.

sivadeilra:
I think that's a good place to start. I do appreciate the requirements for stability, and not casually taking dependencies on crates that might change.

Let's start with this, and I'll see if I can't nudge the safe-transmute process along.  If that ever converges, then we can revisit this.

Thanks!

KodrAus:
We’ve now added `zerocopy` support through #538

PR #500, opened 2020-12-09
prepare for 0.6.6 release

KodrAus:
This updates `rand` to `0.7.x` in our `0.6.x` minor of `uuid` to simplify some uncertain licensing with Fuchsia.


KodrAus:
It looks like we'd rather not patch previous versions of `uuid`, so I'll go ahead and close this one!

Issue #475, opened 2020-06-24
A new version with getrandom?


RazrFalcon:
Is there are plans on making a new release with `getrandom` instead of `rand`?

stjepang:
Ping - I am also interested in a new release for the same reason.

theduke:
I'm also very interested in this, especially for WASM support of `new_v4()`.

kinggoesgaming:
Hi,

I recently switched from full time Linux to full time windows setup. As soon as I am done setting up my dev env I will try to get the PRs merged and all :)

kinggoesgaming:
Ok so I am thinking tomorrow to start looking into this :)


Sorry for the long pause on development

ammgws:
@kinggoesgaming Is #494 a blocker for this?

RazrFalcon:
Any updates?

KodrAus:
We've added `getrandom` support to the `master` branch. I'm backporting now to `0.8.x` so we can push out a new release with it.

KodrAus:
`0.8.2` has now been published :tada:

I gave the diff a good long look to make sure there weren't breaking changes in there, but that's one of the challenges of infrequent releases!

PR #503, opened 2020-12-21
Un-breaking-change Uuid::new_v4

CAD97:

# Description

#447 changed from using `rand::thread_rng` to using `getrandom` in `Uuid::new_v4`. This also changed the return type from `Uuid` to `Result<Uuid, getrandom::Error>`. This PR reverts the signature to the simpler `new_v4() -> Uuid`.

# Motivation

This signature is much simpler to use, and avoids a breaking change. `getrandom` is _highly_ unlikely to fail, and previously we used `thread_rng` here, which also panics if it fails to initialize from the OS entropy source. If `getrandom` fails, it is highly unlikely that any program creating v4 UUIDs has any reasonable recovery other than to abort, as the system is in a broken state.

If users absolutely need to recover in this situation, they can call `getrandom` first themselves to make sure that their system is working, or generate the bytes themselves and create the UUID from those bytes.

Additionally, actually wrapping `getrandom::Error` in `uuid::Error` comes with its own fun set of problems, described in #502.

# Tests

N/A

# Related Issue(s)

#475: this undoes the breaking change to `Uuid::new_v4`, thus making the requested publish a patch update

Closes #502: alternate approach to the same TODO

### DIFF
--- a/src/v4.rs
+++ b/src/v4.rs
@@ -4,10 +4,10 @@ impl Uuid {
     /// Creates a random UUID.
     ///
     /// This uses the [`getrandom`] crate to utilise the operating system's RNG
-    /// as the source of random numbers. If you'd like to use a custom generator,
-    /// don't use this method: generate random bytes using your custom generator
-    /// and pass them to the [`uuid::Builder::from_bytes`][from_bytes] function
-    /// instead.
+    /// as the source of random numbers. If you'd like to use a custom
+    /// generator, don't use this method: generate random bytes using your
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks @CAD97! It looks like we haven't published a release with the dependency on `getrandom`. I think treating it as infallible like you're doing here is the best approach.

I'm not sure what state our main branch is at the moment so I've also created a `0.8.x` branch off `0.8.1` to backport this to.

KodrAus (approved):

PR #505, opened 2020-12-21
Improve deserialization error message

### LINKED ISSUE #499: Error message when trying to parse a random string as a UUID is not helpful
**Describe the bug**

When trying to deserialize something that should contain a UUID, but contains some other string that doesn't happen to be 32 or 36 characters long, one gets this error message from serde:

`invalid length: expected one of [36, 32], found <something>`

This is very unhelpful.

**Expected behavior**

I would expect the error to state that UUID parsing failed. The fact that it failed at a length check would be nice to have as a detail, but is far less important.

**Other**

I think a good fix for this would be to change the serde code to wrap the UUID parsing error in `format!("UUID parsing failed: {}", e)` or something like that. If you want, I'll open a PR.

jplatte:
**I'm submitting a** refactor

# Description

Improves the deserialization error message by including that a UUID was attempted to be parsed.

Before: `invalid length: expected one of [36, 32], found 11`
After: `UUID parsing failed: invalid length: expected one of [36, 32], found 11`

# Motivation

It was not easy to tell from the previous error message what went wrong when trying to deserialize a larger object containing a UUID.

# Tests

I added a unit test.

# Related Issue(s)

Closes #499

### DIFF
--- a/src/serde_support.rs
+++ b/src/serde_support.rs
@@ -13,7 +13,6 @@ use crate::prelude::*;
 use core::fmt;
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
-#[cfg(feature = "serde")]
 impl Serialize for Uuid {
     fn serialize<S: Serializer>(
         &self,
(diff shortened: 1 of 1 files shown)


kinggoesgaming (approved):
This looks good to me

PR #504, opened 2020-12-21
Merge master into 0.8.x

KodrAus:
This PR merges the current `master` branch into `0.8.x` along with a new commit that reverts the breaking change to `from_guid`. I'll give the diff another look over, but this should let us push a new `0.8.x` release with the last year's worth of effort.

Closes:

- #475

PR #506, opened 2020-12-22
rework description of wasm crate features

KodrAus:
As noted in #504, our description of the WebAssembly features isn’t quite accurate anymore now that we’ve shifted from `rand` to `getrandom`. In either case we enable the `js` feature of `getrandom` which just pulls in both `stdweb` and `js-sys` unconditionally.

r? @CAD97

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,14 +43,11 @@
 //! * `serde` - adds the ability to serialize and deserialize a UUID using the
 //!   `serde` crate.
 //!
-//! You need to enable one of the following Cargo features together with
-//! `v3`, `v4` or `v5` feature if you're targeting `wasm32-unknown-unknown`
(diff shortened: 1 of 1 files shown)


CAD97:
This looks good to me. The only thing I'm potentially worried about is

> Windows XP and stdweb are, as of `getrandom` version 0.2.1, no longer supported. If you require support for either of these platforms you may add a dependency on `getrandom = "=0.2.0"` to pin this version. -- [[The Rand book](https://rust-random.github.io/book/update-0.8.html#dependencies)]

I think the recommended behavior (at least the one `rand` uses) is that we would just use `getrandom` directly without exposing or setting any features, and downstream would add a dependency on `getrandom` to customize platform support. Something to consider for future breaking versions.

PR #507, opened 2021-01-04
Use TryInto to avoid unsafe.

### LINKED ISSUE #488: Unsafe is no longer necessary as of rustc 1.34
This code (and another identical line in the same file) uses unsafe to convert a `&[u8]` to a `&[u8; 8]`: https://github.com/uuid-rs/uuid/blob/f177d20151f23a8beb9a75df40c736b27dead0d1/src/lib.rs#L360

From Rust 1.34 and on (before that, TryFrom wasn't stable), the following code compiles to the exact same assembly as the code above:
```rust
self.as_bytes()[8..16].try_into().unwrap()
```

Tested with rustc 1.34 and 1.45.2 on godbolt.org, with `-O`.

The `try_into` version is simpler and doesn't use unsafe, so it'd probably be a good idea to switch to that once support for Rust 1.32/33 is dropped.

Gaelan:

# Description

Previously, to_fields and to_fields_le used unsafe to convert a &[u8] into a &[u8; 8]. Now that we're only supporting Rust versions where TryInto is stable, we can use try_into().unwrap() instead, making uuid entirely safe Rust.

In release mode, the compiler detects that the slice will always be the correct size, so try_into can never fail. Thus, the unwrap
is optimized out and we end up with the exact same assembly as the unsafe block.

Godbolt output showing the resulting assembly: https://godbolt.org/z/nWxT6W

# Motivation

Makes UUID entirely safe Rust.

With this PR,

# Tests

All existing tests pass. Doesn't add any new functionality, so that should be sufficient. Assuming the Godbolt test is consistent with what happens in the context of the larger crate, this shouldn't change the resulting binary at all.

# Related Issue(s)

Closes #488.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -198,7 +198,7 @@ mod v5;
 #[cfg(all(windows, feature = "winapi"))]
 mod winapi_support;
 
-use crate::std::{convert::TryInto, fmt, str};
+use crate::std::{convert, fmt, str};
 
 pub use crate::error::Error;
 
@@ -390,7 +390,8 @@ impl Uuid {
         let d3 =
             u16::from(self.as_bytes()[6]) << 8 | u16::from(self.as_bytes()[7]);
 
-        let d4: &[u8; 8] = self.as_bytes()[8..16].try_into().unwrap();
+        let d4: &[u8; 8] =
+            convert::TryInto::try_into(&self.as_bytes()[8..16]).unwrap();
         (d1, d2, d3, d4)
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -198,7 +198,7 @@ mod v5;
 #[cfg(all(windows, feature = "winapi"))]
 mod winapi_support;
 
-use crate::std::{fmt, str};
+use crate::std::{convert::TryInto, fmt, str};
```
kinggoesgaming, line 201: For consistency, only import modules

kinggoesgaming (changes requested):

src/lib.rs
Gaelan: Ok, switched to just importing std::convert, and using `convert::TryFrom::try_from()` instead of `.try_from()` (which, AAUI, you can't do without importing the trait itself).

kinggoesgaming: Yup it’s done intentionally for traits to prevent any type’s implicit methods interfering...

KodrAus: @kinggoesgaming Hmm, they'll only conflict with other traits imported with conflicting methods. If a type has an inherent method then it'll be chosen over any trait ones.

kinggoesgaming (approved):

PR #511, opened 2021-01-16
Allow access of inner Uuid for adapters

### LINKED ISSUE #482: Allow access of inner Uuid for Hyphenated, Simple, Urn
This PR was recently merged to sqlx, which allows for Uuid support for the mysql driver. Not only can this encode/decode to the basic Uuid type, but we'd also included Hyphenated in there too: https://github.com/launchbadge/sqlx/pull/536

This is especially helpful, because databases have a number of ways the UUID may be serialized back, e.g. the query may select hex-encoded hyphenated strings -- an increasingly common scenario, as these UUIDs may be pulled out of string fields, or stored json. We can encode + decode to a Hyphenated (or Simple, or Urn) easily enough, but it will force us to decode from a Hyphenated string if we ever want it as raw Bytes, even though those Bytes are already part of the struct.

Any way to access the inner is fine. Not sure of the name -- a `From<Hyphenated> for Uuid` would be okay, as would `.to_uuid()`, `.to_inner()`, etc. I'd like it to keep to the style of the crate, hence why I'm opening an issue to ask for help instead of submitting a PR on this fairly simple change.

**Is it blocking?**
No. It's easy to work-around, just overly complex and poor performance.

**Describe alternatives you've considered**
I'd considered a a `Deref` or `AsRef` trait as a way to get a `&Uuid` that references the interior Bytes, if you want to avoid changing ownership of those Bytes, and it'd be easy enough to `.to_owned` or `.clone` that if you want. It'd also make it easy to write generics that take `Deref<Target = Uuid>` or `AsRef<Uuid>`, so you could pass a Uuid, Simple, Hyphenated, or Urn to them and they'd all work the same way. An `Into<Uuid>` would be nice for similar reasons. My point against this is that a `[u8; 16]` can `Copy` just fine, as it's only 2-words long.

hecsalazarf:
**I'm submitting a** feature.

# Description
Trait implementations on the adapters:
* `From<Adapter> for Uuid`
* `AsRef<Uuid> for Adapter`
* `Borrow<Uuid> for Adapter`

An `Adapter::into_inner` method that takes self and returns a `Uuid`.

# Motivation
Open issue #482

# Tests
All tests pass.

# Related Issue(s)
Closes #482

### DIFF
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -1001,14 +1001,14 @@ macro_rules! impl_adapter_from {
 
         impl<$a> AsRef<Uuid> for $T<$a> {
             #[inline]
-            fn as_ref(&self) -> &$a Uuid {
+            fn as_ref(&self) -> &Uuid {
                 self.0
             }
         }
 
         impl<$a> Borrow<Uuid> for $T<$a> {
             #[inline]
-            fn borrow(&self) -> &$a Uuid {
+            fn borrow(&self) -> &Uuid {
                 self.0
             }
         }
--- a/src/adapter/mod.rs
+++ b/src/adapter/mod.rs
@@ -12,7 +12,7 @@
 //! Adapters for various formats for UUIDs
 
 use crate::prelude::*;
-use crate::std::{fmt, str};
+use crate::std::{borrow::Borrow, fmt, str};
 
 #[cfg(feature = "serde")]
 pub mod compact;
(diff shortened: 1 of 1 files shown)


src/adapter/mod.rs
```diff
@@ -1001,14 +1001,14 @@ macro_rules! impl_adapter_from {
 
         impl<$a> AsRef<Uuid> for $T<$a> {
             #[inline]
-            fn as_ref(&self) -> &$a Uuid {
```
hecsalazarf: Removed lifetimes to support rust 1.34.0

KodrAus (approved):

PR #512, opened 2021-01-29
Allow v3, v4, v5 features on wasm32-unknown-unknown

Expyron:
I'm submitting a feature

# Description
The v3, v4, v5 features did not work on `wasm32-unknown-unknown` in a no_std context.

For v3 and v5, the `md5` and `sha1` crates are no_std if their default features are disabled, so they work on `wasm32-unknown-unknown`.

For v4, `getrandom` can work on any architecture if the root crate registers a custom implementation.

This is not a breaking change, as the other architectures (including `stdweb` and `wasm-bindgen`) are not impacted.

 
# Motivation
I wanted to use this crate (especially the v4 feature) on a `wasm32-unknown-unknown` context that does not have `stdweb` nor `wasm-bindgen`.

# Tests
The test suite unfortunately does not run on `wasm32-unknown-unknown`.
The v3, v4, v5 features work properly, as I was able to use them from a dependent crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ optional = true
 version = "0.2.0"
 
 [dependencies.md5]
+default-features = false
 optional = true
 version = "0.7"
 
@@ -68,6 +69,7 @@ optional = true
 version = "1.0.56"
 
 [dependencies.sha1]
+default-features = false
 optional = true
 version = "0.6"
 
--- a/README.md
+++ b/README.md
@@ -46,13 +46,16 @@ various pieces of functionality:
 * `serde` - adds the ability to serialize and deserialize a `Uuid` using the
   `serde` crate.
 
-You need to enable one of the following Cargo features together with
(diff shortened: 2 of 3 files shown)


KodrAus:
Thanks for the PR @Expyron!

I’m a little unsure about committing to `getrandom` here by guaranteeing a registered implementation will work, but otherwise this seems totally reasonable to me!

Expyron:
I also want to point out that `getrandom` will [refuse to compile](https://github.com/rust-random/getrandom/blob/master/src/lib.rs#L213) on an unsupported target if a custom function is not provided, with a clear error message:
> error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets

That error message would be preferable to the current behavior, which is that `Uuid::new_v4()` is not defined at all.

lastmjs:
Is there anything holding this up? I'm starting to develop on DFINITY's Internet Computer, which is a custom WebAssembly environment, and it doesn't like like I can use UUID easily until this is fixed

kinggoesgaming (approved):

PR #521, opened 2021-07-24
Update Uuid::new_v4 documentation to reflect `Try` no longer being implemented

Tehnix:
**I'm submitting a(n)** other (documentation fix)

# Description
The current example at https://docs.rs/uuid/0.8.2/uuid/index.html#examples to generate a UUIDV4 currently does not compile, given that `Uuid` no longer implements `Try`.

This PR updates the example to match the one given in the README of this repo, along with enabling testing for it again to ensure it doesn't go out of date. I've run `cargo test --all-features` to ensure that the tests do indeed pass.

Finally, we also hide the configuration flags, `#[cfg(feature = "v4")]`, from the example, since it should not be copied by the user of the crate, if they wanted to use the example.

# Motivation
I was looking to generate a UUID V4 and unfortunately the examples in https://docs.rs/uuid/0.8.2/uuid/index.html#examples were out-of-date.

# Tests

This enables the doc-test that was already there, which now compiles after updating the code in the example.

# Related Issue(s)
None that I'm aware of

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,12 +92,11 @@
 //!
 //! use uuid::Uuid;
 //!
-//! fn main() -> Result<(), Box<dyn std::error::Error>> {
-//!     #[cfg(feature = "v4")] {
-//!         let my_uuid = Uuid::new_v4()?;
-//!         println!("{}", my_uuid);
-//!     }
-//!     Ok(())
+//! fn main() {
+//! #    #[cfg(feature = "v4")] {
+//!      let my_uuid = Uuid::new_v4();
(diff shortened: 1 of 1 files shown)


kinggoesgaming (approved):
Thank you

Tehnix:
@KodrAus Hmm, is there anything I can do from my end w.r.t. the timeouts? (I don't have permission to the Bors output, so can't see what goes wrong) :/

KodrAus:
@KodrAus r+

KodrAus:
That should do it 😁

PR #525, opened 2021-08-12
add more ci for wasm and embedded targets

KodrAus:
This PR does some more work on our CI to explicitly test wasm through Node and that we build for a proper no-std target. It also uses `cargo-hack` instead of manually listing out feature combinations to test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   os_tests:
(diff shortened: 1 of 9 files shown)


KodrAus:
I've ended up bumping our MSRV to `1.46.0` to support some new dev infrastructure. It would be nice if we could pull in dev dependencies for specific targets to avoid this, but `1.46.0` is now a year old, is available in Linux package managers, and opens up a bunch new const-fn opportunities for the crate.

Issue #469, opened 2020-03-29
alloc no_std support


DianaNites:
**Is your feature request related to a problem? Please describe.**
A project i'm writing is replacing it's std feature with an alloc one,
but uuid only documents std and no_std support.

**Describe the solution you'd like**
An alloc feature to be added, depending only on stuff in `alloc`. Or, if the existing std support actually only depends on alloc, for it to be renamed.

**Is it blocking?**
Kind of?

**Describe alternatives you've considered**
Just use normal no_std uuid

**Additional context**
Documentation of uuid's no_std support seems kind of lacking?, so it's unclear to me what functionality is missing without std/alloc support? It's possible this wouldn't even be useful?

**Other**
The v3 uuid feature depends on the `md5` crate, which itself offers a no_std mode, but defaults to `std`. Uuid does not disable it's default features.

v4(`getrandom`) and v5(`sha1`) are always no_std by default, though.

The `alloc` crate is stable since Rust 1.36

KodrAus:
I don't think `uuid` needs `alloc` for anything so we should be able to work with just about all features except maybe `v4` in `no_std`.

Any libraries we depend on should be `no_std` if they can be so if we're missing any then PRs to fix them up would be welcome!

KodrAus:
In #528 I added some more embedded CI support that ensures the `v1`, `v3`, `v5`, and `serde` features will build on an embedded target. That just makes sure we don't accidentally pull in `std` in any of our dependencies.

I've also added some docs on how to build `uuid` for no-std targets. I think that should just about cover it for this on 🙂

PR #528, opened 2021-08-16
check more features in no-std

KodrAus:
For #469

This checks a few more features against a proper no-std target, to make sure we're not accidentally pulling in `std`.

It doesn't currently test `v4`, because it needs an ambient source of randomness. I _think_ that's possible to do with `getrandom`, but we also have other methods for constructing a random UUID without needing the `v4` feature explicitly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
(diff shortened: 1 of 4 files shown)


KodrAus:
I've replaced our usage of `std::sync::atomic` with the [`atomic` crate](https://github.com/Amanieu/atomic-rs). It looks like it will panic at runtime on platforms that don't support atomic operations. This might not be ideal long-term, but lets us build the `v1` feature without needing to complicate our crate features before the `has_atomic` cfg is stabilized.

Issue #530, opened 2021-09-20
wasm32-unknown-emscripten support


yu-re-ka:
**Describe the bug**

Even though uuid is included with the v4 feature, the new_v4 function doesn't exist.

```
error[E0599]: no function or associated item named `new_v4` found for struct `Uuid` in the current scope
  --> src/core/segment_id.rs:43:11
   |
43 |     Uuid::new_v4()
   |           ^^^^^^ function or associated item not found in `Uuid`
```

**To Reproduce**
1. Add uuid dependency with v4 feature
2. Call new_v4 function
3. Build with target wasm32-unknown-emscripten 

**Expected behavior**

Since emscripten provides the rand functions via the usual C api, uuid should not treat `wasm32-unknown-emscripten` different than `x86_64-unknown-linux-gnu`.

It should use libc random via the rand crate as usual.

**Screenshots**

**Specifications (please complete the following information):**
- Target: wasm32-unknown-emscripten
- Version 0.8.2
- Features Enabled: v4

**Additional context**

https://github.com/uuid-rs/uuid/pull/351

Back then nobody thought about emscripten, so they assumed all wasm32 targets would need workarounds.

**Other**
An exception was already added for wasi:
https://github.com/uuid-rs/uuid/pull/477
But really this should be a specific check for wasm32-unknown-unknown and allow everything else.

PR #531, opened 2021-09-20
Feature/emscripten

yu-re-ka:
**I'm submitting a bug fix**

Fixes https://github.com/uuid-rs/uuid/issues/530


KodrAus:
Ah thanks for fixing this @yu-re-ka!

Issue #422, opened 2019-08-07
uuid!("") literal macro


awakecoding:
I am looking for a way to specify uuid literals in my code without having to do Uuid::parse_str("<literal>").unwrap(); at runtime. A good example for a different use case is the hex-literal crate (https://crates.io/crates/hex-literal) that makes is easy to specify hex literals with the hex!() macro. In this case, I would like to declare a UUID as a string like "cc58f58b-ca41-4678-b9f9-94ba4eeb9e1e" and have it checked for validity at compile-time to avoid the unwrap().

Maybe the best way would be to have the macro define a byte array to use Uuid::from_bytes() under the hood. Any thoughts?

Dylan-DPC:
Hi. Thanks for creating this issue. 

A macro was always part of our plans but we never found time to implement and polish it. We decided to make it a separate repo. You can find it here: https://github.com/uuid-rs/uuid_macro. If you want to contribute to it, let us know and you are free to do so. Thanks.

awakecoding:
@Dylan-DPC thanks, I didn't know someone had begun working on it!

kinggoesgaming:
Just to add on top of @Dylan-DPC's comment:

The eventual plan is to have the `uuid!` macro available directly with `uuid` crate (as a public dependency exposure) post `1.0` release of `uuid`

Hawk777:
The `uuid_macro` crate has been yanked, but as far as I can tell no macro is available in this crate yet. Are there no macros at this point?

kinggoesgaming:
Yes, it's currently yanked. It was only created for name reservation. We currently don't have plans to actually provide the macro until after 1.0 release.

KodrAus:
@QnnOkabayashi That's true. I think it would be great to have a literal macro in the repository that we could look at re-exporting in `uuid` under a feature gate. We could convert this repository into a workspace and implement here if that's a direction you're interested in.

QnnOkabayashi:
@KodrAus Yes, I think this would be great, and I'm happy to deprecate `compiled_uuid` in favor of a built-in solution. What would the process for converting to a workspace look like? I'm also not familiar with how to exactly create feature gates either.

KodrAus:
[...]

PR #543, opened 2021-10-31
Impl UUID macro

QnnOkabayashi:
**I'm submitting a feature**

# Description
Implementing the `uuid!` macro for parsing string literals at compile time.

# Motivation
Very convenient for config files

# Tests
I duplicated the valid str parsing tests for `uuid!()`, and they all compiled. For testing for failure, I checked that invalid inputs don't work:

1. non literal: `uuid!(struct)` fails properly.
2. non string literal: `uuid!(44)` fails properly.
3. invalid string: `uuid!("hello world")` fails properly

I didn't bother to test more extensively then that, do we care?

Also, can someone verify that tokenizing to `::uuid::Uuid::from_bytes([#tokens])` is hygienic?

# Related Issue(s)
https://github.com/uuid-rs/uuid/issues/422


macros/src/lib.rs
```diff
@@ -1,17 +1,106 @@
...
+use std::fmt;
+use syn::spanned::Spanned;
+
+#[cfg(any(feature = "std", test))]
+#[macro_use]
+extern crate std;
+
+#[cfg(all(not(feature = "std"), not(test)))]
+#[macro_use]
+extern crate core as std;
```
QnnOkabayashi, line 13: Required to stop `shared/error.rs` and `shared/parser.rs` from complaining. I just copy-pasted it from `uuid`.

KodrAus:
Thanks for working on this @QnnOkabayashi!

I think we might need to wrap this up in a `macro_rules` macro in `uuid` that can pass the fully-qualified path to the `Uuid` type using the `$crate` metavariable. We could then move the docs to this public macro and make the proc-macro a private implementation detail.

It would also be great to test this using `trybuild` so we can look at diagnostics for invalid inputs, and test that renaming `uuid` or defining a root module called `uuid` doesn’t cause the macro to break.

Is this something you’d be interested in exploring further? If not I’m happy to merge and fill those bits in.

Thanks again! 🙇

QnnOkabayashi:
> I think we might need to wrap this up in a `macro_rules` macro in `uuid` that can pass the fully-qualified path to the `Uuid` type using the `$crate` metavariable. We could then move the docs to this public macro and make the proc-macro a private implementation detail.

Done 👍 
> 
> It would also be great to test this using `trybuild` so we can look at diagnostics for invalid inputs, and test that renaming `uuid` or defining a root module called `uuid` doesn’t cause the macro to break.

Done 👍 

Also since the macro is integrated with the crate now, I was able to use the internal diagnostic information in the `ErrorKind` type to improve the diagnostics. For example, if there's an invalid character, it only points to that character
```
error: invalid character: expected an optional prefix of `urn:uuid:` followed by 0123456789abcdefABCDEF-, found G at 20
  --> benches/macros/invalid_parse.rs:25:44
   |
25 | const _: Uuid = uuid!("F9168C5E-CEB2-4faa-BGBF-329BF39FA1E4");
   |                                            ^
```
If there's an invalid group length, it points to that group
```
error: invalid group length: expected 12, found 8 in group 4
  --> benches/macros/invalid_parse.rs:26:48
   |
26 | const _: Uuid = uuid!("01020304-1112-2122-3132-41424344");
   |                                                ^^^^^^^^
```

KodrAus (approved):
This looks great, thanks @QnnOkabayashi!

I should've merged this one ahead of my changes to our benches. I just renamed some of the files. We probably don't really need benches for the `uuid!` macro, since it should be "free" 🙂

KodrAus:
Besides the merge conflict this should be ready to go now.

QnnOkabayashi:
> Besides the merge conflict this should be ready to go now.

Should be good to go now!

PR #548, opened 2021-11-01
Remove winapi support in favor of examples

### LINKED ISSUE #537: Move winapi support into a new unstable crate
For #113 

Move the `winapi` support into a new unstable crate that's dedicated to UUID conversions:

```rust
pub trait IntoUuid {
    fn into_uuid(self) -> Uuid;
}

pub trait FromUuid {
    fn from_uuid(uuid: Uuid) -> Self;
}
```

We can then implement this trait for `winapi`'s `Guid` and for the `windows-rs` `Guid`.

This will move our last remaining unstable public dependency out of the `uuid` crate so it can be stabilized.

KodrAus:
Closes #537 

Instead of worrying about a conversion library right now this just removes the `winapi` support and adds some examples of creating UUIDs from GUIDs. We can't be certain that all GUIDs can be converted into UUIDs without flipping fields so our `to_guid` and `from_guid` methods don't seem like they could ever be guaranteed to always be correct.


KodrAus:
I've also fixed up my broken changes to the `v1::Context` so it goes back to its old implementation of always incrementing the counter, but will wrap on the max value representable by the UUID, not the max value in a `usize`.

PR #550, opened 2021-11-01
Add optional arbitrary impl

### LINKED ISSUE #514: impl Arbitrary for Uuid to enable easy integration of cargo-fuzz
It's common to have _obejcts_ that use `Uuid`. And whenever someone wants to use `rust-fuzz`, all the components
of the _object_ need to implement/derive the [Arbitrary trait](https://docs.rs/arbitrary/1.0.0/arbitrary/trait.Arbitrary.html)

Having an implementation for `Arbitrary` for `Uuid`

Something like this would be a starting point:

```rust
impl Arbitrary<'_> for Uuid {
    fn arbitrary(u: &mut Unstructured<'_>) -> ArbitraryResult<Self> {
        let b = u.bytes(16)?;
        Ok(Uuid::from_bytes(b.try_into().unwrap()))
    }
}
```

However, I assume it will panic most of the time. So, I'm not sure what's the best way to go about it.

KodrAus:
Closes #514 

Adds an optional implementation of the stable `Arbitrary` trait to `Uuid` so it can be used for fuzzing.

We always give you a valid V4 UUID from the input, unless there aren't enough bytes.

I've also streamlined some of our templates a bit, since I find myself removing most of the scaffolding when writing PRs and issues.


CONTRIBUTING.md
```diff
@@ -13,7 +13,7 @@ Thank you for your interest in contributing to the Uuid Project!
 * [Helpful Links](#helpful-links)
 
 For any questions, please make a post on [users.rust-lang.org][u-r-l-o], post
-on [uuid-rs mailing list] or join our [gitter] channel.
+on [uuid-rs mailing list].
```
kinggoesgaming, line 16: I think we aren't using the mailing list anymore.

kinggoesgaming (changes requested):
Other than removing the mailing list (since you touched that file), everything looks good

CONTRIBUTING.md
kinggoesgaming: I think instead we can unit GitHub Discussions at some point in the future.

KodrAus: I've used Discussions elsewhere and they've been good for keeping issues focused on actionable things. I'd be keen to use them here in `uuid` too

kinggoesgaming (approved):

KodrAus:
Thanks for the review @kinggoesgaming! I've been hammering at these last mile stabilization tasks so have made a huge pile of basically unreviewable changes. I was going to open a PR that prepares for our `1.0.0-alpha` release and lists everything out so we can review the whole end result.

kinggoesgaming:
Should be good!

PR #552, opened 2021-11-01
Add support for parsing UUIDs between braces

KodrAus:
# Description

This PR lets us format and parse UUIDs that are surrounded by braces:

```
{6d93bade-bd9f-4e13-8914-9474e1e3567b}
```

These are common in Windows, and correspond to the `B` format specified for C#'s GUIDs. They also have a format surrounded by parenthesis but I've never seen that one used myself in practice.

# Related Issue(s)

Closes #513

PR #539, opened 2021-10-31
Fix typo in v4.rs

Takashiidobe:
# Description

Fixes a typo in V4.rs `retreive` -> `retrieve`.

PR #520, opened 2021-05-27
Update mod.rs

tshepang:
missing word

**I'm submitting a(n)** (bug fix|deprecation|feature|refactor|removal|other)

# Description

# Motivation

# Tests

# Related Issue(s)

PR #510, opened 2021-01-15
Add impl AsRef<[u8]> for Uuid

hecsalazarf:
**I'm submitting a** feature.

# Description
Implement `AsRef<[u8]>` for `UUID`.

# Motivation
Open issue #472

# Tests
All existing tests pass.

# Related Issue(s)
closes #472

PR #553, opened 2021-11-02
Prepare for 1.0.0-alpha.1 release

KodrAus:
Part of #113 

This PR gets us ready for an alpha release of `uuid` `1.0` that can be used to ensure everything is good in downstream consumers. Ideally we won't make any API changes before actually stabilizing `1.0` proper, but some things may come up that need additional breaking alpha releases.

This release includes a _huge_ amount of work from a lot of contributors. I'll try call out breakage here as well as in release notes once we're ready to merge so everyone knows what to expect.

# Changes since the last release

https://github.com/uuid-rs/uuid/compare/0.8.2...main

# Contributions since the last release

- #510 
- #482 
- #511 
- #507 
- #518 
- #519 
- #512 
- #520 
- #521 
- #525 
- #526 
- #527 
- #528 
- #535 
- #536 
- #539 
- #538 
- #540 
- #541 
- #542 
- #544 
- #545 
- #546 
- #543 
- #547 
- #548 
- #549 
- #550 
- #552 
- #554 
- #558 
- #560 
- #562 
- #563 
- #564 
- #565
- #566

# Highlights

Parsing and formatting methods are now much faster, and work in `const` too! On my i9 9900K Windows desktop, the difference looks something like this:

Case                    | `1.0.0-alpha.1`    | `0.8.3`
----------------------- | ------------------ | ------------------
parse_invalid_character | 40 ns/iter (+/- 0) | 39 ns/iter (+/- 1)
parse_invalid_group_len | 47 ns/iter (+/- 1) | 52 ns/iter (+/- 5)
parse_invalid_groups    | 58 ns/iter (+/- 1) | 58 ns/iter (+/- 0)
parse_invalid_len       | 57 ns/iter (+/- 2) |  6 ns/iter (+/- 0)
parse_nil               | 16 ns/iter (+/- 0) | 50 ns/iter (+/- 4)
parse_nil_hyphenated    | 17 ns/iter (+/- 0) | 59 ns/iter (+/- 2)
parse_random            | 16 ns/iter (+/- 0) | 42 ns/iter (+/- 1)
parse_random_hyphenated | 18 ns/iter (+/- 0) | 51 ns/iter (+/- 2)
parse_urn               | 18 ns/iter (+/- 0) | 51 ns/iter (+/- 1)

You can go one step further and look at the [`uuid-simd`](https://github.com/Nugine/uuid-simd) library for vectorized UUID parsing. It's _fast_!

You've got two options for compile-time UUIDs:

You can create UUIDs at compile time using the `uuid!` macro:

```rust
#[macro_use]
extern crate uuid;

use uuid::Uuid;

const ID: Uuid = uuid!("67e55044-10b1-426f-9247-bb680e5fe0c8");
```

Enable the `macro-diagnostics` crate feature to get better diagnostics from `uuid!` using a procedural macro.

# Breaking changes

## Relocation of `adapter::compact`

This module can now be found in `serde::compact`. It can be combined with `#[serde(with)]` to serialize a `Uuid` using a more compact representation. We originally moved to this representation directly, but discovered it comes with its own drawbacks that aren't suitable for all formats.

## Infallible constructors

The following methods have been changed to accept a `&[u8; N]` instead of a `&[u8]`, making them infallible instead of returning a `Result`:

- `Uuid::from_fields`
- `Uuid::from_fields_le`
- `Uuid::new_v1`
- `Builder::from_fields`
- `Builder::from_fields_le`

## Infallible `get_variant`

The `Uuid::get_variant` method now returns a `Variant` instead of an `Option<Varaint>`, because it's not actually possible for it to ever be `None`.

## `Uuid::to_timestamp` is now `Uuid::get_timestamp`

The `Uuid::to_timestamp` method has been renamed to `Uuid::get_timestamp` to make it more consistent with `Uuid::get_version` and `Uuid::get_variant`.

## Changes to formatting methods

The following methods that produce formatting adapters have been renamed:

- `Uuid::to_hyphenated` -> `Uuid::hyphenated`
- `Uuid::to_simple` -> `Uuid::simple`
- `Uuid::to_urn` -> `Uuid::urn`

The following types have been removed:

- `HyphenatedRef`
- `SimpleRef`
- `UrnRef`

The following methods have been changed to return a `&<AdapterType>` instead of an `<AtapterType>Ref`:

- `Uuid::to_hyphenated_ref` -> `Uuid::as_hyphenated`
- `Uuid::to_simple_ref` -> `Uuid::as_simple`
- `Uuid::to_urn_ref` -> `Uuid::as_urn`

## `Builder` method consistency

The `Builder::build` method has been renamed to `Builder::into_uuid`, to complement the `<AdapterType>::into_uuid` methods. It also gets an equivalent `Builder::as_uuid` method.

## `Version` and `Variant` are non-exhaustive

The `#[non_exhaustive]` attribute has been added to `Version` and `Variant`. There are already active proposals for new UUID versions, so these are likely to continue evolving in the future.

## Removed `winapi` support

The `Uuid::to_guid` and `Uuid::from_guid` methods have been removed. This was done for two reasons:

- We need to avoid unstable `winapi` types in the stable `uuid` public API.
- Whether or not GUID fields need to be flipped to convert into a UUID depends on the source of the GUID, so we can't really offer a single pair of methods to convert a GUID into a UUID.

There are some examples in the repository root that demonstrate how to convert GUIDs into UUIDs as a replacement.

## Building with `--all-features`

Now that `uuid` includes unstable features, if you're building with `--all-features` (such as in package manager scenarios), you'll also need to pass `RUSTFLAGS="--cfg uuid_unstable"`, or you'll end up with compilation errors. If this strategy becomes problematic for users we can change unstable features to silently no-op instead of cause compilation failures if the corresponding cfg is not also supplied. Please reach out if that affects you!

# Stability commitment

Once `uuid` releases `1.0` we'll try to stay on that version "forever". There won't be a `2.0` unless there's a very compelling need for it. That means you should feel safe depending on `Uuid` in the public API of your libraries if you want to.

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -1,7 +1,18 @@
 [package]
-name = "uuid_macro"
-version = "0.0.0"
+name = "uuid-macro-internal"
+version = "1.0.0-alpha.1"
 edition = "2018"
+authors = [
+    "QnnOkabayashi"
+]
+categories = [
+    "data-structures",
+    "no-std",
+    "parser-implementations",
+    "wasm"
+]
+description = "Private implementation details of the uuid! macro."
+documentation = "https://docs.rs/uuid"
 
 [lib]
 proc-macro = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,11 +30,10 @@ homepage = "https://github.com/uuid-rs/uuid"
 name = "uuid"
 readme = "README.md"
 repository = "https://github.com/uuid-rs/uuid"
-version = "0.8.1" # remember to update html_root_url in lib.rs
+version = "1.0.0-alpha.1" # remember to update html_root_url in lib.rs
 
 [package.metadata.docs.rs]
-features = [ "serde", "slog", "v1", "v3", "v4", "v5" ]
-default-target = "x86_64-pc-windows-msvc"
+features = ["serde", "arbitrary", "slog", "v1", "v3", "v4", "v5"]
 
 [package.metadata.playground]
 features = ["serde", "v1", "v3", "v4", "v5"]
@@ -54,7 +53,7 @@ repository = "uuid-rs/uuid"
 [features]
 default = ["std"]
 std = []
-macro-diagnostics = ["uuid_macro"]
+macro-diagnostics = ["uuid-macro-internal"]
 
 v1 = ["atomic"]
 v3 = ["md-5"]
@@ -98,7 +97,7 @@ optional = true
 version = "0.9"
 
 # Public: Re-exported
-[dependencies.uuid_macro]
+[dependencies.uuid-macro-internal]
 path = "macros"
 optional = true
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ To get started with generating random UUIDs, add this to your `Cargo.toml`:
 
 ```toml
 [dependencies.uuid]
-version = "0.8"
+version = "1"
 features = ["v4", "fast-rng"]
 ```
 
@@ -113,6 +113,11 @@ flag through your environment to opt-in to unstable `uuid` features:
 RUSTFLAGS="--cfg uuid_unstable"
 ```
 
+## Minimum Supported Rust Version (MSRV)
+
+The minimum supported Rust version for `uuid` is documented in
+CI. It may be bumped in minor releases as necessary.
+
 ## References
 
 * [Wikipedia: Universally Unique Identifier](     http://en.wikipedia.org/wiki/Universally_unique_identifier)
@@ -120,7 +125,7 @@ RUSTFLAGS="--cfg uuid_unstable"
 
 [`wasm-bindgen`]: https://github.com/rustwasm/wasm-bindgen
 
-[`Uuid`]: https://docs.rs/uuid/0.8.1/uuid/struct.Uuid.html
+[`Uuid`]: https://docs.rs/uuid/1.0.0-alpha.1/uuid/struct.Uuid.html
 
 ---
 # License
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,8 +35,13 @@
 //! Add the following to your `Cargo.toml`:
 //!
 //! ```toml
-//! [dependencies]
-//! uuid = { version = "0.8", features = ["v4"] }
+//! [dependencies.uuid]
+//! version = "1.0.0-alpha.1"
+//! features = [
+//!     "v4",                # Lets you generate random UUIDs
+//!     "fast-rng",          # Use a faster (but still sufficiently random) RNG
+//!     "macro-diagnostics", # Enable better diagnostics for compile-time UUIDs
+//! ]
 //! ```
 //!
 //! When you want a UUID, you can generate one:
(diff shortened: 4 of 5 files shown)


macros/Cargo.toml
```diff
@@ -1,7 +1,18 @@
 [package]
-name = "uuid_macro"
-version = "0.0.0"
+name = "uuid-macro-internal"
+version = "1.0.0-alpha.1"
 edition = "2018"
+authors = [
+    "QnnOkabayashi"
```
kinggoesgaming, line 6: Should this include atleast one `uuid-rs` team member?

Unless I misunderstand how initial ownership for crates.io works

kinggoesgaming (dismissed):

macros/Cargo.toml
KodrAus: This field’s just an acknowledgement of the primary authors so I thought I’d put just @QnnOkabayashi here for now as the current sole contributor to this piece.

kinggoesgaming: ah yes.. ok. 

clearly I don't understand basic ownership. 😅

KodrAus:
Just cross-posting a comment from Reddit on the change to `serde`:

> For some serialization formats, byte array is more efficient per element than seq. Tuple is similar to seq, except it has known length. For example, consider a format where each element has one byte overhead, where tuple would thus cost 2*16 bytes instead of 1+16 bytes for byte array.

https://www.reddit.com/r/rust/comments/qksulj/comment/hizps4r/

Dylan-DPC:
How about bumping the edition to 2021 (in a separate PR)? I know there
won't be much changes but I think its better to keep it to the latest
edition.

mitsuhiko:
I think the serde change is not positive. It's untypical, generally serde hasn't been used like this and it's going to break a lot of formats. I'm also not convinced that u128 is a better solution here. Does the length information cause a problem in practice that it's worth considering going out of the way to do something untypical here?

KodrAus:
@mitsuhiko I've opened #557 that we can use to try come to some resolution. We won't be shipping any builds with a changed `serde` implementation until we do.

KodrAus:
The change to the default serialization format has been reverted, so `Uuid` will continue to serialize using byte slices as it currently does.

KodrAus:
I think we've been through and over the feedback that came in from the community, so the last thing to do is a final pass over the API to make sure nothing's been missed. I've done this myself, but would appreciate any more eyes on it! You can use:

```
RUSTDOCFLAGS="--cfg uuid_unstable" cargo doc --all-features --open
```

to explore the complete public API.

macros/Cargo.toml
KodrAus: Ah yes, it's separate from the ownership rights on crates.io 🙂

KodrAus:
cc @kinggoesgaming how is everything looking to you now?

kinggoesgaming:
@KodrAus Following the current activity on repo, this should be good.

Will have a look in the morning through the entire firebase just to see if there is anything missed out.

kinggoesgaming:
@KodrAus I think all is good for the first alpha release.

KodrAus:
Thanks for reviewing this @kinggoesgaming!

PR #554, opened 2021-11-02
Fixed `uuid!("")` macro diagnostics when using `urn:uuid` prefix.

QnnOkabayashi:
**I'm submitting a** bug fix

# Description
Fixed an bug where the `uuid!("")` macro would point to the wrong characters when using the `urn:uuid:` prefix or wrapping with the curly braces (`"{...}"`).

# Related Issue(s)
none

### DIFF
--- a/shared/parser.rs
+++ b/shared/parser.rs
@@ -36,22 +36,24 @@ pub fn parse_str(mut input: &str) -> Result<[u8; 16], Error> {
 
     let mut start = 0;
     // Check for a URN prefixed UUID
-    if len == 45 && input.starts_with(URN_PREFIX) {
-        input = &input[URN_PREFIX.len()..];
-        start += URN_PREFIX.len();
+    if len == 45 {
+        if let Some(stripped) = input.strip_prefix(URN_PREFIX) {
+            input = stripped;
+            start = URN_PREFIX.len();
+        }
     }
     // Check for a Microsoft GUID wrapped in {}
-    else if len == 38 && input.starts_with("{") && input.ends_with("}") {
-        input = &input[1..input.len() - 1];
-        start += 1;
+    else if len == 38 {
+        if let Some(stripped) =
+            input.strip_prefix('{').and_then(|s| s.strip_suffix('}'))
+        {
+            input = stripped;
+            start = 1;
+        }
     }
     // In other cases, check for a simple or hyphenated UUID
     else if !len_matches_any(len, &[36, 32]) {
-        return Err(ErrorKind::InvalidLength {
-            expected: ExpectedLength::Any(&[36, 32]),
-            found: len,
-        }
-        .into());
+        return Err(Error::length(ExpectedLength::Any(&[36, 32]), len));
     }
 
     // `digit` counts only hexadecimal digits, `i_char` counts all chars.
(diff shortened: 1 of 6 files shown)


KodrAus:
Nice catch!

Looks like we’ve got a couple test failures here. I think the `start` variable is expected to be fixed, so we might need to track the index to report errors at from a different starting point?

We also just need to make sure the parser will never panic. Maybe that should go in a comment at the top of the function?

QnnOkabayashi:
> Nice catch!
> 
> Looks like we’ve got a couple test failures here. I think the `start` variable is expected to be fixed, so we might need to track the index to report errors at from a different starting point?
> 
Fixed the indexing error that was causing usize overflow
> We also just need to make sure the parser will never panic. Maybe that should go in a comment at the top of the function?

I also found out that it panicked when a wide char was present, so I had to add some support in the macro that fixed this.

Also, I added some util functions for creating `Error`s to clean up the parsing algorithm a bit.

shared/parser.rs
```diff
@@ -127,21 +124,17 @@ pub fn parse_str(mut input: &str) -> Result<[u8; 16], Error> {
...
                 _ => {
-                    return Err(ErrorKind::InvalidCharacter {
-                        expected: "0123456789abcdefABCDEF-",
-                        found: input[i_char..].chars().next().unwrap(),
-                        index: i_char,
-                        urn: UrnPrefix::Optional,
-                    }
-                    .into());
+                    // let found = input[i_char..].chars().next().unwrap();
+                    // let found = char::from(chr);
```
KodrAus, line 136: Did we mean to keep these comments here?

tests/ui/compile_fail/invalid_parse.rs
```diff
@@ -25,5 +25,18 @@ const _: Uuid = uuid!("F9168C5E-CEB2-4faa-BBF-329BF39FA1E4");
...
+const _: Uuid = uuid!("urn:uuid:F9168C5E-CEB2-4faa-B2cBF-32BF39FA1E4");
+const _: Uuid = uuid!("{F9168C5E-CEB2-4faa-B0a75-32BF39FA1E4}");
+
+const _: Uuid = uuid!("{F9168C5E-CEB2-4faa-B6BF-329Bz39FA1E4}");
+
+// group 0 has invalid length
+const _: Uuid = uuid!("67e550-4105b1426f9247bb680e5fe0c");
+
+const _: Uuid = uuid!("504410岡林aab1426f9247bb680e5fe0c8");
+const _: Uuid = uuid!("504410😎👍aab1426f9247bb680e5fe0c8");
```
KodrAus, line 38: 😎 👍

shared/parser.rs
QnnOkabayashi: Nope that was an accident

KodrAus:
That is much easier to follow. Do we have any test cases for the `parse_str` function itself that includes wide characters?

QnnOkabayashi:
> That is much easier to follow. Do we have any test cases for the `parse_str` function itself that includes wide characters?

Not yet. Also how optimized is the parsing algorithm? The more I look at it, the more it bothers me how clunky it feels, and I'd like to refactor it if you're not opposed.

KodrAus:
> I'd like to refactor it if you're not opposed.

I think it's had one or two rounds of optimization applied in its lifetime, but nothing like a complete rewrite as far as I remember. If you'd like to spend some time on it that sounds fantastic! I don't mind if you just want to keep pushing changes to this PR here, or if we merge and start another.

PR #558, opened 2021-11-03
Make conversion methods consistent across the API

KodrAus:
This PR removes the explicit `Ref` types for different formats in favor of casting a `&Uuid`. This is safe to do because the format types are `#[repr(transparent)]` and `Copy`.

I've also made the naming conventions consistent for all methods that produce and return `Uuid`s and enhanced the docs.


KodrAus:
@kinggoesgaming This basically changes our format adapters to be a single type. We make a reference to these types from a UUID when we want to avoid copying. This is a safe transmute because they’ve got the same ABI, implement `Copy`, and don’t implement `Drop`.

Issue #561, opened 2021-11-14
SIMD accelerations


Nugine:
**Motivation**
According to my benchmarks, SIMD parser can be 5x~9x faster than generic implementation and SIMD formatter can be 10x faster than generic implementation.

<https://github.com/Nugine/uuid-simd/blob/main/benches/main.rs>
<https://github.com/Nugine/uuid-simd>

**Solution**
Add SIMD functions and a feature `simd`. 
When `simd` and `std` are enabled, use runtime cpu feature detection to dispatch the functions. 
When `simd` is enabled, use conditional compilation to dispatch the functions. 
Otherwise, use the fallback functions.

**Alternatives**
SIMD accelerations don't have to be built-in features. An extension crate (`uuid-simd`) is also good for users who care about performance.

**Is it blocking?**
No.

**Anything else?**
`Uuid::try_parse` is slightly slower than my impl. I'm testing why this happens. I'll submit a PR if I find the reason.

```
parse/fallback/simple         | 27.286 ns   
parse/fallback/hyphenated     | 28.057 ns
parse/uuid-latest/simple      | 40.266 ns
parse/uuid-latest/hyphenated  | 41.445 ns
```

Edit:
uuid-simd link: https://docs.rs/uuid-simd/0.5.0/uuid_simd/

KodrAus:
I think in `uuid` it would be great to offer something based on `std::simd` once that becomes available, but the burden of maintaining specialized implementations for different extensions might be a bit much for us. We could add a `simd` feature that pulled in a library of vectorized parsing and formatting functions to replace `uuid`'s internals with. We should also call out `uuid-simd` in the README!

KodrAus:
Since we mention `uuid-simd` in the README I think we can go ahead and close this one for now. The next opportunity we'll have to action it will be sometime in the future, so we can either re-open or create a fresh issue then.

Issue #551, opened 2021-11-01
Inconsistent inlining?


kinggoesgaming:
https://github.com/uuid-rs/uuid/blob/main/src/fmt.rs#L19-L62

I have seen inconsistent inlining for the various `Uuid` `fmt` traits. I am not sure that is by design, since I did not review the changes when they were made. Just wanted to bring it up, since we are that close to `1.0` release.

kinggoesgaming:
Actually also exists for the adapters as well, inside the macro rules, further down the file

KodrAus:
Ah, this is quite possibly not by design 😄 

We can add these inlines at any time, but it would be great to have a look and see which ones are worthwhile.

kinggoesgaming:
Considering that the release is still looking a bit away (looking at you,  #557), I'll to bench inlines and inlines tomorrow and get a PR open according

KodrAus:
I think our inlines are probably pretty accurate now after all the optimization work done in the last few months. I'll go ahead and close this one, and any specific issues that come up can create fresh ones.

PR #562, opened 2021-11-14
Improve `try_parse` and `encode` performance

Nugine:
I'm submitting a refactor

# Description
This PR improves the performance of `Uuid::{try_parse, parse_str}` and `{Simple, Hyphenated, Braced, Urn}::{encode_lower, encode_upper}`

| bench                   | before | after     |
| ----------------------- | ------ | --------- |
| encode_hyphen           | 25 ns  | **15 ns** |
| encode_simple           | 23 ns  | **15 ns** |
| encode_urn              | 29 ns  | **16 ns** |
| hyphenated              | 43 ns  | 37 ns     |
| simple                  | 39 ns  | 32 ns     |
| urn                     | 46 ns  | 37 ns     |
| parse_invalid_character | 72 ns  | 63 ns     |
| parse_invalid_group_len | 81 ns  | 73 ns     |
| parse_invalid_groups    | 94 ns  | 90 ns     |
| parse_invalid_len       | 88 ns  | 86 ns     |
| parse_nil               | 46 ns  | **26 ns** |
| parse_nil_hyphenated    | 47 ns  | **28 ns** |
| parse_random            | 46 ns  | **26 ns** |
| parse_random_hyphenated | 47 ns  | **28 ns** |
| parse_urn               | 49 ns  | **29 ns** |

# Related issues
#551
#561

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -112,9 +112,7 @@ impl Uuid {
     #[inline]
     pub fn as_hyphenated(&self) -> &Hyphenated {
         // SAFETY: `Uuid` and `Hyphenated` have the same ABI
-        unsafe {
-            &*(self as *const Uuid as *const Hyphenated)
-        }
+        unsafe { &*(self as *const Uuid as *const Hyphenated) }
     }
 
     /// Get a [`Simple`] formatter.
@@ -131,9 +129,7 @@ impl Uuid {
     #[inline]
     pub fn as_simple(&self) -> &Simple {
         // SAFETY: `Uuid` and `Simple` have the same ABI
-        unsafe {
-            &*(self as *const Uuid as *const Simple)
-        }
+        unsafe { &*(self as *const Uuid as *const Simple) }
     }
 
     /// Get a [`Urn`] formatter.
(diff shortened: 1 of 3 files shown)


src/fmt.rs
```diff
@@ -187,58 +179,109 @@ const LOWER: [u8; 16] = [
...
-        .expect("found non-ASCII output characters while encoding a UUID")
+#[inline]
+fn encode_simple<'b>(
+    src: &[u8; 16],
+    buffer: &'b mut [u8],
+    upper: bool,
+) -> &'b mut str {
+    const LEN: usize = 32;
+    let buf = &mut buffer[..LEN];
+    unsafe {
```
KodrAus, line 232: Maybe we could move the SAFETY comment to outside of the `unsafe` block here to also cover the `ptr::write` call?

The `as_mut_ptr` call isn't unsafe so we could also pull that one out.

KodrAus, line 232: Maybe we could move the SAFETY comment to outside of the `unsafe` block here to also cover the `ptr::write` call?

The `as_mut_ptr` call isn't unsafe so we could also pull that one out.

KodrAus (approved):
Thanks for working on this @Nugine! I just had one nit-picky comment around SAFETY comments on the `unsafe` blocks.

It would be great to come back through and add some docs describing how we parse the various shaped UUIDs. At the moment I think a casual passer-by might have a hard time following along. I'm happy to come back and do that later though unless you'd like to write it up.

PR #565, opened 2021-11-16
some small doc cleanups

KodrAus:
For #553 

This is just a final pass over our crate docs, fixing up any little things I notice along the way.


QnnOkabayashi (approved):
We should explicitly mention that `uuid!()` works on string literals :)

KodrAus:
Thanks @QnnOkabayashi!

PR #566, opened 2021-11-16
Provide uuid macro without dependencies

### LINKED ISSUE #556: Macro construction without `syn`
**Motivation**
Macros are notoriously slow at compile time to include, it'd be nice to remove the tradeoff for users

**Solution**
The `uuid` macro can be implemented with const fn ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=333ef1bacbf891afb185280d7abc7220)). This makes the decision easier for users - just move the call to a const context.

If we wish, we can also have the `uuid` macro to force the parsing to be const.

Nugine:
I'm submitting a feature

# Description

https://github.com/uuid-rs/uuid/issues/556#issuecomment-959352417

# Related Issue(s)

resolves #556

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -241,7 +241,6 @@ mod rng;
 
 mod external;
 
-#[cfg(feature = "macros")]
 #[macro_use]
 mod macros;
 #[doc(hidden)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,7 +70,7 @@
 //!
 //! Other crate features can also be useful beyond the version support:
 //!
-//! * `macros` - adds the `uuid!` macro that can parse UUIDs from string literals at compile time.
+//! * `macro-diagnostics` - enhances the diagnostics of `uuid!` macro.
 //! * `serde` - adds the ability to serialize and deserialize a UUID using
 //!   `serde`.
 //! * `arbitrary` - adds an `Arbitrary` trait implementation to `Uuid` for
@@ -243,8 +243,9 @@ mod external;
 
 #[macro_use]
 mod macros;
+
 #[doc(hidden)]
-#[cfg(feature = "macros")]
+#[cfg(feature = "macro-diagnostics")]
 pub extern crate uuid_macro;
 
 use crate::std::convert;
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,3 +1,37 @@
+macro_rules! define_uuid_macro {
+    {$(#[$doc:meta])*} => {
+        $(#[$doc])*
+        #[cfg(feature = "macros")]
+        #[macro_export]
+        macro_rules! uuid {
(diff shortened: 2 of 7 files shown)


KodrAus:
cc @QnnOkabayashi

src/macros.rs
```diff
@@ -1,3 +1,37 @@
...
+            ($uuid:literal) => {{
+                const OUTPUT: $crate::Uuid = match $crate::Uuid::try_parse($uuid) {
+                    Ok(u) => u,
+                    Err(_) => {
+                        // here triggers const_err
+                        // const_panic requires 1.57
+                        #[allow(unconditional_panic)]
+                        let _ = ["invalid uuid representation"][1]; 
+
+                        loop {} // -> never type
```
KodrAus, line 25: 😂

QnnOkabayashi (commented):
At this point, we should change the feature gate to something like `"macro-diagnostics"`. I also personally think that it should be on by default and require opting-out, since nice error messages should be the default. Thoughts?

src/macros.rs
Nugine: Tricky, but it works :joy:
<https://docs.rs/static_assertions/1.1.0/src/static_assertions/const_assert.rs.html#52-57>

KodrAus:
A `macro-diagnostics` feature sounds like a good one to me :+1: Ideally we would have it on by default, but unfortunately for a non-trivial subset of users any additional dependencies will be pushed back against. That's one of our main reasons for a non-proc-macro version, and is also what landed us with a `fast-rng` feature that pulls in `rand`. I think there's an alternative approach to default features though that we can take: recommend a set of features to enable at the top of the README and in the crate-level docs:

```toml
[dependencies.uuid]
version = "1"
features = [
    "v4",                # allow generating random UUIDs
    "fast-rng",          # use a faster (but still sufficiently random) algorithm
    "macro-diagnostics", # support more detailed diagnostics in the `uuid!` macro
]
```

It's not on by default, but it's recommended by default.

src/lib.rs
```diff
@@ -70,7 +70,7 @@
 //!
 //! Other crate features can also be useful beyond the version support:
 //!
-//! * `macros` - adds the `uuid!` macro that can parse UUIDs from string literals at compile time.
+//! * `macros` - enhances the diagnostics of `uuid!` macro.
```
KodrAus, line 73:
```suggestion
//! * `macro-diagnostics` - enhances the diagnostics of `uuid!` macro.
```

KodrAus (commented):
Thanks @Nugine! This looks good to me. I just had one suggestion on a missed feature rename in the docs.

src/lib.rs
Nugine: It's a miss when I rebased the commits :joy:

PR #567, opened 2021-11-17
Fix up publishing for uuid

KodrAus:
It turns out the `/shared` approach for the parser and error doesn't work with `cargo publish`, because it won't include sources from outside the crate directory, nor from within any nested directories.

For now, just duplicate the source, since there isn't too much of it. We should try come up with a better approach in the future that doesn't require adding dependencies to UUID.


kinggoesgaming:
I was wondering what took so long to publish the version to crates.io

KodrAus:
😅

macros/COPYRIGHT
```diff
@@ -0,0 +1,8 @@
+The Uuid Project is copyright 2013-2014, The Rust Project Developers and
+copyright 2018, The Uuid Developers.
```
kinggoesgaming, line 2: I think the macro itself does not need The Rust Project Developers copyright?

src/parser.rs
```diff
@@ -92,13 +88,154 @@ impl Uuid {
...
     pub const fn try_parse(input: &str) -> Result<Uuid, Error> {
-        match imp::try_parse(input) {
+        match try_parse(input) {
             Ok(bytes) => Ok(Uuid::from_bytes(bytes)),
             Err(_) => Err(Error(ErrorKind::Other)),
         }
     }
 }
 
+#[inline]
```
kinggoesgaming, line 98: I am not sure about this inline...

while in a `const` context this works fine... however in non-`const` this is bound to create a lot of bloat

kinggoesgaming (changes requested):

kinggoesgaming:
Found a couple of things... not necesarily blockers... but nice to address while we are at it

src/parser.rs
KodrAus: These inlines don't seem to be doing much for us, so I removed them :+1:

Nugine: These inlines have observable effects on the performance.

If you want to avoid code bloat, just write a wrapper function with `#[inline(never)]`

https://rust-lang.github.io/rust-clippy/master/index.html#missing_inline_in_public_items

kinggoesgaming (approved):

src/parser.rs
kinggoesgaming: This is an interesting point. 

@KodrAus Thoughts?

KodrAus: @Nugine I didn’t observe any difference in removing these two specifically, even at the level of optimization we’re at now where small perturbations make a big difference, so opted for fewer inlines.

Nugine (approved):

KodrAus:
We may actually just want to let these two parsers diverge a bit so the diagnostics parser is optimized specifically for that case.

PR #573, opened 2021-11-19
Follow C-GETTER convention

vilgotf:
Renames getters to follow [C-GETTER](https://rust-lang.github.io/api-guidelines/naming.html#c-getter)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ use uuid::{Uuid, Version};
 
 let my_uuid = Uuid::parse_str("67e55044-10b1-426f-9247-bb680e5fe0c8")?;
 
-assert_eq!(Some(Version::Random), my_uuid.get_version());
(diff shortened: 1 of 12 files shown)


kinggoesgaming:
I will let @KodrAus comment on this as he was the one to last modify it.

KodrAus:
Thanks for the PR @vilgotf! These are arguably better names, but I was cautious about churning them so opted to keep them with their existing `get_` prefixes.

KodrAus:
Since we've published a `1.0` release candidate I think the ship has basically sailed on these kinds of changes now. Unless there's some critical bug or soundness issue to fix I think we should avoid making any breaking changes, and treat the API we've got now as fixed with a small window for patching any critical issues that might arise.

Thanks for the PR though @vilgotf! :bow:

PR #581, opened 2022-01-17
Update hashes to v0.10 and change sha-1 to sha1

newpavlov:
`sha1` got transferred to RustCrypto and should be preferred over `sha-1`, which will be deprecated after v0.11 release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,13 +88,13 @@ version = "0.5"
 [dependencies.md-5]
 default-features = false
 optional = true
-version = "0.9"
+version = "0.10"
 
 # Private
-[dependencies.sha-1]
+[dependencies.sha1]
 default-features = false
 optional = true
-version = "0.9"
+version = "0.10"
 
 # Public: Re-exported
 [dependencies.uuid-macro-internal]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ macro-diagnostics = ["uuid-macro-internal"]
 v1 = ["atomic"]
 v3 = ["md-5"]
 v4 = ["rng"]
-v5 = ["sha-1"]
+v5 = ["sha1"]
 
 js = ["getrandom", "getrandom/js"]
 


KodrAus:
Thanks @newpavlov! I hadn't noticed the crate transfer before. I think we originally moved to the rust-crypto implementation because the original was unmaintained. That doesn't seem to be the case now. Do you have a sense of how these two implementations differ now? Is the rust-crypto version faster on some platforms? (I got that impression just from a very quick scan including some arch-specific code).

newpavlov:
Apart from the crate name, sha1 v0.10 and sha-1 v0.10 are completely identical. `sha-1` will be completely deprecated in favor of `sha1` on the next breaking release, but for v0.10 we plan to release identical patch releases for both of them.

KodrAus:
Ah I was more wondering about the difference between `sha1` and `sha1_smol` 🙂 I had a bit of a dig through things and for `sha1`, I think there's very little reason for `uuid` specifically to prefer the more industrial rust-crypto `sha1` over the zero-dependency `sha1_smol`, since they're both now maintained. They perform roughly identically (I couldn't try the asm-based implementation yet because I'm on MSVC), and we don't interact with the wider infrastructure rust-crypto makes available through its traits.

So I think we'll take a similar approach that we have with RNG and `rand` vs `getrandom`, where we'll quietly use `sha1_smol` by default because it has zero dependencies, but allow a `fast-sha1` with `sha1/asm` that users could opt into if their target platform supports it.

KodrAus:
Ok, I flipped over to my ARM-based MacBook and the asm-based implementation is quite a lot zippier, so I think we'll go ahead with this approach of defaulting to `sha1_smol` and supporting `sha1/asm` through a feature gate, just like we already do with RNG. Thanks for the PR @newpavlov 🙏

newpavlov:
You already have `md-5` in your dependency tree. Adding `sha-1` will not add any additional dependencies.

Also `sha1_smol` is as unmaintained as the old `sha1` implementation. In my opinion, it does not make much sense to revert back to it.

PR #595, opened 2022-04-06
Remove unnecessary AND

clehner:
`& 0x1f` sets the left three bits to zero, but `| 0xe0` sets them back to one. The AND (`&`) can therefore be removed as unnecessary. This is similar to how in the NCS case there is only an AND (no OR).

- 0x1f = 00011111
- 0xe0 = 11100000


KodrAus (approved):
Thanks @clehner! This looks good to me.

PR #606, opened 2022-06-09
Fix some doc links

mbrobbel:
Some doc links were broken (referencing `adapter`). This PR fixes those links and changes some links (in the same file) to shortcut reference links.


KodrAus (approved):
Thanks for cleaning this up @mbrobbel!

Issue #609, opened 2022-07-03
Codebase formatting


malobre:
While working on #523 I noticed:
- some files are not correctly formatted (i.e: running `cargo fmt --check` fails),
- a `.rustfmt.toml` file is present,
  it was added in #210 because there was differences in formatting between nightly and stable.
  This might not be needed anymore ?

PR #610, opened 2022-07-11
formatting changes

pintariching:
**I'm submitting a refactor**

# Description
Formatting the codebase a bit by following `cargo fmt --check` recommendations

# Related Issue(s)
#609

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: 1.46.0
+        toolchain: 1.57.0
         override: true
 
     - name: Version features
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,9 +1,0 @@
-max_width = 80
-hard_tabs = false
-tab_spaces = 4
-newline_style = "Unix"
-wrap_comments = true
-normalize_comments = true
-merge_derives = true
-force_explicit_abi = true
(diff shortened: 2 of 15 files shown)


pintariching:
After a bit of formatting `cargo fmt --check` returns a wall of errors
```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `normalize_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `unstable_features = false`, unstable features are only available in nightly channel.
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `normalize_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `unstable_features = false`, unstable features are only available in nightly channel.
```

As stated in #609 I also tried deleting the `.rustfmt.toml` file but oh boy, cargo was not happy and gave me around 60 diffs. Should I refactor it this way?

KodrAus:
Hi @pintariching :wave:

If you'd like to delete the `rustfmt.toml` and let those formatting changes come through that sounds good to me. Those are nightly-only configuration toggles and we should probably just use the stable formatter for this library. We haven't been running formatting in CI (I'd much prefer to just run it on merge) so there will be a big diff, but it can live in a single commit.

Thanks for working on this!

KodrAus:
It looks like our minimum supported Rust version can't build `serde_test` anymore. We've been meaning to bump it to `1.57.0`, which is around 6 months old now, so I think we can do that here.

Issue #523, opened 2021-08-07
UUIDv6–v8 Support


XAMPPRocky:
**Is your feature request related to a problem? Please describe.**
There is a new draft RFC for UUID formats that are more database friendly, it would be nice if these were supported in the `uuid` crate, since the new formats share a lot of the same internals.

**Describe the solution you'd like**
An implementation of UUIDv6, v7, and v8 from the [IETF RFC](https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format).

**Is it blocking?**
No, it would just be cool to have.

-----

**EDIT (@KodrAus):** The RFC is being worked on [here](https://github.com/ietf-wg-uuidrev/rfc4122bis)

KodrAus:
Looks like I've got some reading to do :eyes: These are basically ULIDs are they?

XAMPPRocky:
I hadn't seen ULID before, but looking at the project it seems very similar, UUID has more variety in what kind of timestamp you can use.

VladimirMarkelov:
Today I ran across a project that collects all UUIDv6 prototypes. It lacks Rust implementation :) If you add UUIDv6 support to your library, it would be great

https://github.com/uuid6/prototypes/

VladimirMarkelov:
> I hadn't seen ULID before, but looking at the project it seems very similar, UUID has more variety in what kind of timestamp you can use.
> 
> Looks like I've got some reading to do 👀 These are basically ULIDs are they?

As I understand, UUIDv6 can be sorted and the sorted list would be in the generated time order that make UUIDv6 much more desirable for using as primary key or alike in databases

KodrAus:
Thanks @XAMPPRocky and @VladimirMarkelov. I'll spend some time unpicking the spec to see if it would make sense to share our `v1` timestamp APIs with `v6` or whether there are changes we should make to unify them.

KodrAus:
Ok, based on a really quick reading it looks like our `Timestamp` and `ClockSequence` should already be suitable at least for `v6` and `v7` UUIDs. It already can convert between the original Gregorian format and a standard Unix epoch. I haven't looked at `v8` just yet.

KodrAus:
[...]

PR #611, opened 2022-08-17
refactors Timestamp, Context and ClockSequence, and adds UUIDS v6, v7, v8

rrichardson:
**I'm submitting a feature** 

# Description 
This addresses all requested features in #523

Adds v6, v7 and v8 features and modules. 
It also adds a Timestamp::now()  which does what you might expect and uses std::time::SystemTime. 
I added some basic tests for each, but they could certainly use more tests and more eyes. 

## Breaking changes
This refactors Timestamp to make it more generally applicable to all (time based) UUIDs. 
It also breaks context/sequence out of Timestamp, since that only applies to v1 and v6.  
This changes the API for v1 and v6, to require a Context to be supplied at creation. 
Updates the crate to Edition 2021

## Dependency Changes 
Removes a dependency on atomic and instead uses std::sync::atomic since it now supports the necessary types.  

# Related Issue(s)
#523

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -307,7 +307,7 @@ impl Uuid {
     /// Basic usage:
     ///
     /// ```
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// # use uuid::Uuid;
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
@@ -348,7 +348,7 @@ impl Uuid {
     /// Basic usage:
     ///
     /// ```
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// # use uuid::Uuid;
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
@@ -383,7 +383,7 @@ impl Uuid {
     /// Basic usage:
     ///
     /// ```
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// # use uuid::Uuid;
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
@@ -414,7 +414,7 @@ impl Uuid {
     /// Basic usage:
     ///
     /// ```
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// # use uuid::Uuid;
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
@@ -446,7 +446,7 @@ impl Uuid {
     /// Basic usage:
     ///
     /// ```
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// # use uuid::Uuid;
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
@@ -515,7 +515,7 @@ impl Builder {
     /// Basic usage:
     ///
     /// ```
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// # use uuid::{Builder, Uuid};
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
@@ -589,7 +589,7 @@ impl Builder {
     ///
     /// ```
     /// # use uuid::Builder;
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn main() -> Result<(), uuid::Error> {
     /// let bytes = [
     ///     0xa1, 0xa2, 0xa3, 0xa4,
     ///     0xb1, 0xb2,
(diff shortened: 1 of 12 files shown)


robjtede:
Just want to say from a lib maintainers perspective, there's basically no good reason that such a widely used crate should need to get bumped to `edition = 2021`. It hurts compatibility for folks on earlier compilers/editions far more than it helps maintainability (compared to 2015->2018 shift).

rrichardson:
@robjtede - Thanks for the input.   The shift to 2021 really wasn't much more than changing the value in the Cargo.toml.  What about it hurts compatibility?

KodrAus (commented):
Thanks for working on this @rrichardson! :bow:

We can't make breaking changes to UUID's API now that it's stable, so I think our best option with the timestamp is to either:

1. Interpret the subseconds field as a counter in v1/v6 UUIDs and as extra precision in v7/v8
2. Make the timestamp an enum that's either a v1/v6 or v7/v8 timestamp.

Neither of these are as ideal as your refactoring, but I think we need to work within the API we've got.

src/v1.rs
```diff
@@ -183,32 +54,49 @@ impl Uuid {
...
+    /// let _uuid = Uuid::new_v1(ts, &context, &[1, 2, 3, 4, 5, 6]);
+    /// ```
+    ///
     /// [`Timestamp`]: v1/struct.Timestamp.html
     /// [`ClockSequence`]: v1/trait.ClockSequence.html
     /// [`Context`]: v1/struct.Context.html
-    pub const fn new_v1(ts: Timestamp, node_id: &[u8; 6]) -> Self {
-        let time_low = (ts.ticks & 0xFFFF_FFFF) as u32;
-        let time_mid = ((ts.ticks >> 32) & 0xFFFF) as u16;
+    pub fn new_v1(
```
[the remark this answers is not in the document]
rrichardson: OK.. I'll make this PR conform to the existing Library/Tests.  I will make a follow up PR with an RFC for my proposed API refactorings re: https://github.com/uuid-rs/uuid/issues/612

rrichardson:
@KodrAus - Do you know any ways that we can work around the fact that 2 of the CI jobs fail when running the powerset of features (runs the container out of space)  I'm not a Github Actions wizard.    It seems 15 features pushes it over the failing point.

KodrAus:
@rrichardson I'll send a PR through that rejigs that part of CI a bit. The main thing we want to ensure is that each feature works independently, and that they all work together. I'll give you a ping when that's up.

rrichardson:
@KodrAus - Ok.. This should be compatible with the original API.  Note, this API requires some silly hoop-jumping for any versions that use a timestamp but not a counter (e.g. v7  https://github.com/rrichardson/uuid/blob/six-seven-eight/src/v7.rs#L53-L61)

rrichardson:
All checks have passed!  🎊 🍾

src/builder.rs
```diff
@@ -40,7 +40,7 @@ use crate::{error::*, Bytes, Uuid, Variant, Version};
 /// ```
 #[allow(missing_copy_implementations)]
 #[derive(Debug)]
-pub struct Builder(Uuid);
+pub struct Builder(pub Uuid);
```
KodrAus, line 43: Was making this field public intentional?

src/timestamp.rs
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    /// Construct a `Timestamp` from the current time of day
+    /// according to Rust's SystemTime
+    #[cfg(all(feature = "std", not(any(feature = "v1", feature = "v6"))))]
+    pub fn now() -> Self {
+        let dur = std::time::SystemTime::UNIX_EPOCH
+            .elapsed()
+            .expect("Getting elapsed time since UNIX_EPOCH.  If this fails, we've somehow violated causality");
```
KodrAus, line 76: I think this can happen on systems with drifting clocks, maybe more likely to be observed on Windows since its epoch is earlier than the Unix one?

I don't have a great alternative for what we should do, besides documenting that this method will panic if the system time is set earlier than the Unix epoch.

KodrAus (commented):
This is awesome work! Thanks for putting it together, and for sticking through our current API constraints :bow: I just had a few minor comments.

src/timestamp.rs
rrichardson: I think that clock drift shouldn't be a problem.  The Result returned by `SystemTime::elapsed()` is an error if self's time happens **after** now.    In this case, it is checking that that the current system time is later than Jan 1st, 1970.  
I'll update the documentation and the message accordingly.

KodrAus: Yeh, so less drifting clocks, and maybe more likely misconfigured Windows systems, where their epoch is some much earlier date than the Unix one 🤔

src/builder.rs
rrichardson: Yeah.. this was necessary due to moving modules around. actually.. it could probably be pub (crate) if

rrichardson: ok.. I just checked and maybe it's not necessary.   Removed.

src/v6.rs
```diff
@@ -0,0 +1,167 @@
+//! The implementation for Version 1 UUIDs.
+//!
+//! Note that you need to enable the `v1` Cargo feature
```
0xpr03, line 3: I think you meant to say `v6` features
```suggestion
//! Note that you need to enable the `v6` Cargo feature
```

rrichardson:
BTW:   Arbitrary just bumped their patch version which added the use of array::from_fn which was just stabilized in Rust 1.63.  So some of the CI jobs failed because of this.  I've locked Arbitrary to the prior version (1.1.3) until we can bump all CI workers to 1.63

KodrAus:
Thanks for working through this @rrichardson! This is an amazing effort :bow:

PR #613, opened 2022-08-22
Optionally implement `defmt::Format`

falk-h:
**I'm submitting a feature**

# Description
Make `Uuid` work seamlessly with [`defmt`](https://github.com/knurling-rs/defmt).

Potential issues:
- Testing. Not sure how to test this. Does this even need testing?
- `defmt` only supports the latest stable Rust version. Is this a dealbreaker with your MSRV policy?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,11 @@ version = "2"
 optional = true
 version = "1"
 
+# Public: Used in trait impls on `Uuid`
+[dependencies.defmt]
+optional = true
+version = "0.3"
+
 # Public (unstable): Used in `zerocopy` derive
 # Unstable: also need RUSTFLAGS="--cfg uuid_unstable" to work
 # This feature may break between releases, or be removed entirely before
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,7 +81,10 @@ version = "2"
 optional = true
 version = "1"
 
-# Public: Used in trait impls on `Uuid`
+# Public (unstable): Used in trait impls on `Uuid`
+# Unstable: also need RUSTFLAGS="--cfg uuid_unstable" to work
+# This feature may break between releases, or be removed entirely before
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -81,6 +81,11 @@ version = "2"
 optional = true
 version = "1"
 
+# Public: Used in trait impls on `Uuid`
+[dependencies.defmt]
```
KodrAus, line 85: Since the `defmt` library is unstable, we'll need to jump through a few extra hoops to support it directly. If you follow the example of `zerocopy` we'll need to add a l[ine like this to our `lib.rs`](https://github.com/uuid-rs/uuid/blob/main/src/lib.rs#L215-L216) to ensure users also opt-in to an unstable feature flag to use `defmt`:

```rust
#[cfg(all(not(uuid_unstable), feature = "defmt"))]
compile_error!("The `defmt` feature is unstable and may break between releases. Please also pass `RUSTFLAGS=\"--cfg uuid_unstable\"` to allow it.");
```

KodrAus (commented):
Thanks for working on this @falk-h! Adding `defmt` support seems reasonable to me, we just need to make sure we add it unstably, since `defmt` may still have any number of breaking changes coming before it stabilizes.

falk-h:
Got it. Should I create a tracking issue as well?

KodrAus (approved):
Thanks @falk-h! This looks good to me. I'll go ahead and create a tracking issue for this one.

falk-h:
Huh. Not sure why the CI fails on Windows but not on Linux. I unfortunately can't reproduce it either, as I don't have a Windows machine available. Any ideas?

KodrAus:
Yeh, it's a strange failure, I'll check it out :eyes:

KodrAus:
Moved to #639

PR #614, opened 2022-08-22
Get doc tests to pass without std and shrink build runs

KodrAus:
For #611 

Our build requirements explode when new features are added. What we really want is just to make sure that each feature works independently, and that they all work together. This PR does that by avoiding using the feature powerset in CI. I've also made sure our doc tests actually work when the `std` feature is disabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,10 +66,10 @@ jobs:
       run: cargo test --all-features --examples
 
     - name: Each version feature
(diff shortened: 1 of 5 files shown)


rrichardson:
It would appear that the `--each-feature` run fails on Windows.  Presumably because the env vars don't evaluate the same way in powershell.

KodrAus:
Thanks for carrying this forwards, I think that's fair to unblock V6-V8 support, I'll look at reducing the duplication asynchronously.

Issue #612, opened 2022-08-17
API Refactoring


rrichardson:
**Motivation**

The UUID Crate has grow organically over the last 5 years.  It's seen many new features, and groups of features added,  and new uuid version support has been added, where convenient.  The Rust ecosystem has evolved significantly in that time.  The implementations in the crate have sort of evolved in different directions,  there is the Builder, which has functions which are general, and it has functions which are specific to a single UUID version (without mentioning that version's name,  see ::from_random_bytes)).  

There are about 15 features in this crate (if you include the newest versions).  IMO, that number should be closer to 4. 

As a result of its growths and offshoots in various areas, the crate is a bit of a mess, with many inconsistencies and some characteristics that make it look a bit dated. 

**Solution**

I propose we refactor the crate in preparation for a new major release.  
The basic idea would be this: 
1. We remove all version-specific features, and provide 3 top-level, unified APIs. Both APIs can build any Uuid version, which tools they have to build said versions is up to the included features.   The top level APIs would be:   
     1. A high-level builder API that can build all version
     2. A lower-level, module-based API which exposes consistent functions for every version of uuid.  There would also be version-specific functions where needed.  Where each implementation is in its own module, e.g. 
         `uuid::v1::new(...)`
      3. The Uuid struct, which would not know how to construct itself directly, outside of parsing/converting Uuids in other formats.
2. The only Cargo features would be ones that are specific to target platform restrictions (e.g. wasm, embedded, nostd, fast-random vs /dev/random etc) 
3. We put all stateful and random generation things inside a GeneratorContext (or whatever we want to name it) This way, if people wish to use their own RNGs or clock-sequences, or whatever, it can be tucked neatly away inside that. 
4. Timestamp would stay, but would be abstracted away by the high-level functions, people shouldn't have to generate their own time if they don't want to.  (if they disable std, then they would have to bring their own time functions and construct their own timestamps)

If people are OK with this basic arrangement, then I will start working on an RFC.

KodrAus:
Thanks for the insights @rrichardson! I'd be keen to see what an alternative greenfield API might end up looking like. I'd also be keen to see what real benefits that API offers over the current organization. The current design is that:

- `Builder` can construct any UUID version using whatever tools you have to produce them. This is currently true for v3, v4, and v5 UUIDs where users want to plug in their own implementations for randomness and SHA1/MD5 hashing.
[...]

PR #615, opened 2022-08-25
Uuid 2.0 RFC

rrichardson:
**RFC proposal** 

Rendered: https://github.com/rrichardson/uuid/blob/rfc-app-2/docs/rfcs/uuidgen-refactor.md

# Description
This is the core of my proposal for a refactor of the Uuid crate.  Fundamentally, it doesn't change how the existing APIs are used a whole lot, it introduces a TimeSource and changes the ClockSequence to SequenceSource (where a sequence can either be monotonic or random).  

The biggest changes are:  
1. Removal of features for versions. Features for 3rd party functionality (rng, std, etc) remain. 
2. The addition of (and, IMO a focus on) the UuidGen and corresponding Builder struct.  

# Related Issue(s)
#612

### DIFF
--- a/docs/rfcs/uuidgen-refactor.md
+++ b/docs/rfcs/uuidgen-refactor.md
@@ -1,0 +1,133 @@
+
+# Uuid API 2.0
+
+
+### UuidGenBuilder and Uuid Version-Specific Refactoring
+
+The goal of this doc is to lay out a refactored API for a potential Major Version release (read: Not necessarily backwards compatible, though it could be)
+
+The three primary motivations are to create a new concept called UuidGen, which is, as you might guess, a Uuid Generator.  The second reason is to decouple the lower-level API components to make them more suitable for the 7 Uuid versions that we support. The third reason, similar to the second, is to future proof the API and make it easy to implement future, unknown Uuid standards.
+
+In this new scheme, we would eliminate version-specific features.  All versions would be constructable by the API by default.  However, the functionality needed to construct Uuids lives in structs which may be gated by features.  E.g. std, rand, thread-rng, system-rng, atomics, etc.
+
+#### UuidGen / Builder
+
+The UuidGen and Builder types will incorporate **type-state** methods to safely construct a version-specific Uuid Generator which can generate new Uuids simply by calling `next()`
+
+The motivation for this is three fold:
+
+1. People rarely want to create 1 uuid, usually it's millions or more.
+2. Presently, through either the Uuid struct or through the Builder, there are a fair amount of machinations required to create a single Uuid. I hope to demonstrate that one can make a generator with the same amount of effort.
+3. Being synchronous, this generator can also implement an `Iterator` interface, which opens up a world of possibilities for "free".
+
+The implementation of this would involve quite a few structs, most of which represent a partially constructed, version-specific Uuid generator.
+e.g.:
+
+```rust
+   let bldr: UuidV7GenBuilderNoClockNoSeq = UuidGenBuilder::new_v7();
+   let bldr: UuidV7GenBuilderNoSeq<SystemTimeSource> = bldr.with_system_time();                  // assumes feature="std"
+   let bldr: UuidV7GenBuilder<SystemTimeSource, rand::ThreadRng> = bldr.with_thread_local_rng(); // assumes feature="thread-rng"
+   let gen: UuidV7Gen<SystemTimeSource, rand::ThreadRng> = bldr.finish();
+   let u1: UuidV7 = gen.next();
+
+   // or for `no features`
+
+   struct MyTime {};
+   impl TimeSource for MyTime {
+   	fn next_unix_time() -> (u64, u32) {
+	  ...
+	}
+   }
+   struct MySequentialGen {};
+   impl SequenceSource for MySequentialGen {
+        fn next_u64() -> u64 {
+	  ...
+	}
+   }
+   let bldr: UuidV1GenBuilderNoClockNoSeq = UuidGenBuilder::new_v1();
+   let bldr: UuidV1GenBuilderNoClock<MyTime> = bldr.with_seq_source(MySequentialGen {});
+   let gen: UuidV1Gen<MyTime, MySequentialGen> = bldr.with_time_source(MyTimeSource {}).finish();
+   let u2 = gen.next();
+
+   // Note that all of these types are for illustration purposes really a user would just do
+
+   let gen = UuidV1GenBuilder::new_v1()
+       .with_seq_source(MySequentialGen {})
+       .with_time_source(MyTimeSource {})
+       .finish()
+
+```
+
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks for writing this @rrichardson! I'll spend some proper time reading through the RFC, but wanted to set the expectation early that since we committed to a stable `uuid` API that other libraries could depend on publicly we'd need a really compelling reason to make any breaking changes. That reason might be within the text already.

docs/rfcs/uuidgen-refactor.md
```diff
@@ -0,0 +1,222 @@
...
+```rust
+trait TimeSource {
+    type Output;
+    fn next(&mut self) -> Self::Output;
+}
+
+// Note that SequenceSource is parameterized over TimeSource::Output because some implementations might want to track the previous Clock output to see if it was a duplicate before increasing their counter.
+trait SequenceSource<T: TimeSource> {
+    type Output;
+    fn next(&mut self, tm: <T as TimeSource>::Output) -> Self::Output;
```
joshtriplett, line 118: Rather than making this part of `SequenceSource`, what about requiring that `TimeSource::Output` implements `Eq`, and then deciding inside the UUID library whether to check for duplicates or not?

I'm also wondering to what extent it makes sense for `TimeSource::Output` to be parameterized, without having some notion of precision with which to turn it into the component needed for a UUID? Or is the expectation that the library will nonetheless rely on specific output types rather than being generic?

joshtriplett:
:+1: for the proposal to have feature flags only for dependencies. As someone with crates that indirectly depend on uuid, I would appreciate being able to prune dependencies. (Please consider having all of those feature flags not part of `default`.) But eliminating APIs without eliminating dependencies doesn't seem sufficiently high-value to be worth the maintenance costs.

docs/rfcs/uuidgen-refactor.md
rrichardson: I definitely like the idea of removing any awareness of `TimeSource` from `SequenceSource`.   However, whether or not the `SequenceSource` checks the time for duplicates isn't really a version-specific thing, it's a preference of the implementation.  If we could pass in some notion of "there was a conflict" into `SequenceSource::next()` that would be fine by me.  Maybe an `Option<bool>`

Regarding Output precision:   My assumption was that the precision required would be version-specific, so the `Output` would always be parameterized by each `UuidV#` Implementation.   The only exception to this, though, would be UuidV8, which violates this rule.   

TBH, this GenBuilder API was inspired by the idea that one could use a builder to make a highly tuned UuidV8 implementation by choosing how many bits they wanted for seconds, frac-seconds, random and sequence.  However, there might be better/safer ways to accommodate that level of flexibility in the API design.  (I was going to add the precision/size check at runtime in the Builder construction)

rrichardson:
> Thanks for writing this @rrichardson! I'll spend some proper time reading through the RFC, but wanted to set the expectation early that since we committed to a stable `uuid` API that other libraries could depend on publicly we'd need a really compelling reason to make any breaking changes. That reason might be within the text already.

Much of this proposal could be tacked on without changing the existing API.   
However, there are a couple of "bugs" in the existing API that have become obvious when implementing v6-8.   All of these issues I bring up have workarounds, but at the expense of making the API more convoluted. 

Firstly, one can't create a Timestamp without creating an `impl ClockSequence`.  IMO, this is an unnecessary conflation of APIs.  It specifically makes no sense for Uuidv7, which doesn't leverage any clock sequence, it only uses unix-time + random. 

Secondly,  the Traits and timestamps assume too much.  They are fixed at  u64, u32, and u16. This, again, makes certain version implementations a bit awkward and they don't compose well. We should be able to use a SequenceGenerator to generate 128 bit random numbers to serve the v7 and v8 use cases.

Thirdly,  things like `Uuid::get_timestamp()` and `get_version()` do a version switch or throw a runtime error.  This pattern, IMO, is screaming that it should be part of a trait, and simply be version-specific impls, so there is no way to get it wrong. 

Fourthly, (and less important) the `Uuid` struct makes it too easy to use accidentally use Uuid versions interchangeably, and this again, is enforced with a runtime check, when it could easily be enforced by the compiler. 

Lastly,  I think we need to refactor how features are done, because, IMO, it is a bit of a mess.   If we do this, we do need to make sure it's done in a way that is backwards compatible,  or at least define an evolutionary path.

docs/rfcs/uuidgen-refactor.md
joshtriplett: @rrichardson That does seem like perhaps more flexibility than necessary.

And yeah, passing in "there was a conflict" seems like an improvement, at least.

rrichardson:
Since this is a large API change, even for a v2.0.   What are your thoughts on putting the new implementation under a new module (called `next` or whatever.   This way people could evolve towards the 2.0 approach..  Then the next step would be to make it part of the main implementation, and put the v1 API under the `old` module.  
We could build a compatibility layer with the old approach.

kinggoesgaming:
> Since this is a large API change, even for a v2.0.   What are your thoughts on putting the new implementation under a new module (called `next` or whatever.   This way people could evolve towards the 2.0 approach..  Then the next step would be to make it part of the main implementation, and put the v1 API under the `old` module.  
> 
> We could build a compatibility layer with the old approach. 

I support this approach. Allows experimentation without affecting current users.

KodrAus:
Thanks for writing this up @rrichardson. At first, this seems like a lot of machinery to generate UUIDs, and I'm not sure that splitting UUIDs into separate types by version is necessarily the best user experience. It seems like the main problem we've got is that the responsibilities of our `Builder` and `Uuid` types are effectively reversed. On `Uuid`, we have high-level methods like [`now_v7()`](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.now_v7) that depend on crate features being enabled, and on `Builder` we have low-level methods like [`from_unix_timestamp_millis`](https://docs.rs/uuid/latest/uuid/struct.Builder.html#method.from_unix_timestamp_millis) that don't depend on any crate features or non-libcore APIs. If the roles were reversed and `Uuid` was *just* low-level construction and formatting it would make considering different strategies for constructing them easier. I somewhat regret that I didn't see the opportunity to pull APIs off `Uuid` and onto the `Builder` when first introducing it, because it would also make it possible to release major versions of the library that all re-exported the same `Uuid`.

I'd welcome any experimentation to build a better API as an external library, but I think the bar for heavy refactoring at this point is high. The end-user pay-off really needs to justify shifting things around, especially when the basic APIs have remained the same for many years.

KodrAus:
One major version I would be keen to do now was one that simply re-exported the entire current API, but made the `v1` and `timestamp` modules private.

KodrAus:
Just coming back in through some triage. At this stage we're not planning any breaking changes to `Uuid`. Its API will remain as-is for as long as we can possibly get away with it. It's been good to explore ways things could be improved, but realistically I don't think we're going to be refactoring. Thanks for having taken the time to write this all up though!

PR #621, opened 2022-09-09
Fixing clippy warning on dead code

gents83:
unix_to_rfc4122_ticks is marked as unused from clippy warnings - I've added the attribute to not display the warning, but if you think it has to be deleted, better to delete it then :)


KodrAus:
Thanks for the PR @gents83! This code is actually used, but only for version 1 and 6 UUIDs, so it does need an attribute. I've fixed it up in [my current PR](https://github.com/uuid-rs/uuid/pull/625/files#diff-df69eeff48c2663ed59b31541c6c396ffecfed985b395daadafda019a524f75cR128) that moves some things around based on the new v6-v8 support that was recently added.

Thanks again for looking into this!

PR #642, opened 2022-10-20
Fix incorrect reference to `try_parse`'s error messages in its doc comment

nick96:
**I'm submitting a** bug fix

# Description

The doc comment for `try_parse` says "if `try_parse` fails it won't generate very useful error messages". It should instead by `parse_str`. Sorry for the tiny PR, it just made me scratch my head for a moment when I came across it!

# Related Issue(s)

None

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -71,7 +71,7 @@ impl Uuid {
     /// hyphens.
     ///
     /// This function is similar to [`parse_str`], in fact `parse_str` shares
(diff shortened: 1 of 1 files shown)


KodrAus:
Hi @nick96 :wave: I think the docs here are actually right, but maybe aren't worded so well. The `try_parse` method will only give you very basic error info, but works in constants, whereas the `parse_str` method gives you more detailed user-facing error info, but can't be used in constants.

Issue #643, opened 2022-11-01
Uuid 1.2x requires alloc


lulf:
**What happened?**

After the commit c60dd940, the `extern crate alloc` dependency was added, which requires any no_std environment to use an allocator to use uuid.

1. Depend on uuid with default-features false
2. Use uuid in project
3. Observe error message requiring allocator

**What were you expecting?**

Maybe guarding 'alloc' behind a feature flag?

KodrAus:
Hm, it shouldn't be necessary for `uuid` to depend on `alloc` at all. I think this `extern crate` should be removed.

KodrAus:
Thanks for the report @lulf :bow: And for the benefit of the doubt that this was actually intentional, and not just a very silly oversight. I'll get this released and look into why [our embedded target in CI](https://github.com/uuid-rs/uuid/blob/main/.github/workflows/ci.yml#L146-L171) didn't seem to catch it.

lulf:
Thanks for fixing it ☺️

PR #644, opened 2022-11-01
Guard alloc requirement behind alloc feature flag

lulf:
Proposed fix for #643


KodrAus:
As noted in #643, this was actually a mistake and not something we want to do deliberately. Thanks for looking into it @lulf

Issue #649, opened 2022-12-30
 Implement FromParam trait for Uuid from the Rocket web framework crate


LMH01:
**Motivation**
I would like to use the `Uuid` struct in a rocket request handler. For that `Uuid` needs to implement the trait `FromParam`.

**Solution**
Add a new feature called "rocket" that will enable an implementation of `FromParam` for `Uuid`.

**Alternatives**
I could use a string instead and check if the string represents a `Uuid` inside the function but doing it this way will result in an unclean error message when this parsing fails. It would be more clean if the function isn't even run when the request guard `FromParam` fails.

**Is it blocking?**
Yes, it blocks me from using it in an request handler. I'm currently rewriting the user authentication to use this Uuid instead of a custom struct. See this issue: [link](https://github.com/LMH01/Acquire_rs_web/issues/1)

**Anything else?**
I have already begun to implement this myself, a pull request will follow.

LMH01:
Closing because this can already be archived by enabling the `uuid` feature inside rocket. See #650 for more information.

PR #650, opened 2022-12-30
Add compatibility with rocket

LMH01:
**I'm submitting a feature**

# Description
This pull requests implements the traits [FromRequest](https://api.rocket.rs/v0.5-rc/rocket/request/trait.FromRequest.html) and [FromParam](https://api.rocket.rs/v0.5-rc/rocket/request/trait.FromParam.html) from the [web framework rocket](https://github.com/SergioBenitez/Rocket) to `Uuid`.

# Related Issue(s)
#649 
#651

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,6 +149,11 @@ package = "wasm-bindgen"
 version = "0.2"
 optional = true
 
+[dependencies.rocket]
+package = "rocket"
+version = "0.5.0-rc.2"
+optional = true
+
 [dev-dependencies.bincode]
 version = "1.0"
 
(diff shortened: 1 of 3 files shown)


LMH01:
#649 and #651 are implemented

LMH01:
@KodrAus
Could you please review this pull request?
It is ready.

LMH01:
Another idea I just had: 
FromRequest currently only passes when the uuid is sent in an http header. I think it would be beneficial if it is also checked if a cookie named uuid exists if the http header does not exist.
This way the user would have two options to transmit the uuid to the request handler and they can decide which method to use.

I will implement that probably tomorrow.

LMH01:
Review is ready again.

a1phyr:
This already exists in `rocket`, but you need the `uuid` feature

Issue #655, opened 2023-01-24
Parsing error mentions incorrect characters as expected


basbossink-ds:
**What happened?**
The error message given in a parsing error when deserializing a readable uuid, suggest to use characters which are illegal.

1. Deserialize a readable uuid with invalid input .eg. `"hello_world"`.
2. An error is created with the message:  
   ~~~        
   "UUID parsing failed: invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-zA-Z], found `h` at 1"
   ~~~

**What were you expecting?**
The expected error message for this case should in my humble opinion be:
~~~
"UUID parsing failed: invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-fA-F-], found `h` at 1"
~~~

**What's the context?**
- Target: rustc 1.66.0 (69f9c33d7 2022-12-12)
- Version: main
- Cargo features enabled: default

PR #656, opened 2023-01-24
Fix error message.

basbossink-ds:
**I'm submitting a** bug fix

# Description
This is a slight adjustment of the error message, since it is contradicting itself.
# Related Issue
#655


KodrAus (approved):
Thanks @basbossink-ds. That's a nice improvement :+1:

PR #673, opened 2023-05-07
Hard deprecate Timestamp::to_unix_nanos

KodrAus:
The `Timestamp::to_unix` method has been deprecated for some time now. I'd like to remove it entirely in a future release. I can't find any actual usages of it, since the method was never useful, so in a `2.x` release of `uuid` it will disappear. Note that since the plan is for `1.x` to re-export `2.x` that means this method will vanish in some future minor release of `1.x`. We don't want to do this often, but the alternative of letting multiple versions of the `Uuid` type float around is worse than having a single deprecated and useless method disappear.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,24 +22,24 @@ jobs:
     strategy:
       matrix:
         exclude:
-        - os: macos-10.15
+        - os: macos-latest
           rust_target: x86_64-gnu
-        - os: macos-10.15
+        - os: macos-latest
           rust_target: x86_64-msvc
-        - os: windows-2019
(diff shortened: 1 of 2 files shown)


nrxus:
@KodrAus  FYI I think this PR incidentally raised the MSRV to 1.57.0. That version came out almost 2 years ago by now so it's not a big deal but it's probably worth changing the `README.md` as it still specifies 1.46.0 as the MSRV which is incorrect.

KodrAus:
Thanks for the note @nrxus! It looks like we actually made the bump way back in #610 but never updated the readme. I'll get that sorted 🙇

Issue #681, opened 2023-06-11
About UUID v7 implementation


Hanaasagi:
This function is used in `Uuid::new_v7()`, it will place the content of `rand_bytes[1]`(12 bits) before `rand_bytes[0]`.

https://github.com/uuid-rs/uuid/blob/6188ecf96c66f06d31d84784c50da182ae5fbb64/src/timestamp.rs#L235

Currently, it has no impact on the generated result because they are both random bytes. Maybe this line should be changed to

```diff
diff --git a/src/timestamp.rs b/src/timestamp.rs
index e49771d..7ec88e6 100644
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -232,7 +232,7 @@ pub(crate) const fn encode_unix_timestamp_millis(millis: u64, random_bytes: &[u8
     let millis_low = (millis & 0xFFFF) as u16;
 
     let random_and_version =
-        (random_bytes[0] as u16 | ((random_bytes[1] as u16) << 8) & 0x0FFF) | (0x7 << 12);
+        (random_bytes[1] as u16 | ((random_bytes[0] as u16) << 8) & 0x0FFF) | (0x7 << 12);
 
     let mut d4 = [0; 8];
```
 

Ref: https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-04#section-5.2

KodrAus:
Well spotted! I don’t think there’s any reason why it should be one way or the other, but the fact that the indexes are reversed might suggest there’s more going on than there actually is.

I’d happily accept a PR that reversed them 🙂

PR #682, opened 2023-06-12
fix: keep the order when filling random bytes

Hanaasagi:
Fix: https://github.com/uuid-rs/uuid/issues/681


KodrAus:
Thanks @Hanaasagi!

PR #686, opened 2023-06-15
Add `borsh` support

### LINKED ISSUE #685: Optional support for borsh
Hi! :wave:

I'm serializing some types using [borsh](https://github.com/near/borsh-rs) but I can't implement traits like [`BorshSerialize`](https://docs.rs/borsh/latest/borsh/ser/trait.BorshSerialize.html) and [`BorshDeserialize`](https://docs.rs/borsh/latest/borsh/de/trait.BorshDeserialize.html) directly on `Uuid` types.

Currently I workaround this by using a newtype that implements those traits (using the `u128` representation for borsh), like
```rust
pub struct Id(Uuid);

impl BorshSerialize for Id {
    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
        BorshSerialize::serialize(&self.0.as_u128(), writer)
    }
}

impl BorshDeserialize for Id{
    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
        Ok(Self(Uuid::from_u128(BorshDeserialize::deserialize_reader(
            reader,
        )?)))
    }
}
```

In practice these newtypes aren't very convenient though, because I end up converting between `Uuid` and `Id` pretty often. I was wondering if it might be useful to upstream this directly onto `Uuid`, like the optional `serde` feature?

I'd be happy to submit a PR if there's any interest in this.

grovesNL:
Fixes #685

I ended up implementing this for my temporary fork anyway, so thought I'd post it here in case there's interest in adding it back to uuid.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Continuous integration
 
 env:
   VERSION_FEATURES: "v1 v3 v4 v5 v6 v7 v8"
-  DEP_FEATURES: "slog serde arbitrary zerocopy"
+  DEP_FEATURES: "slog serde arbitrary borsh zerocopy"
 
 on:
   pull_request:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ version = "1.3.4" # remember to update html_root_url in lib.rs
 rustc-args = ["--cfg", "uuid_unstable"]
 rustdoc-args = ["--cfg", "uuid_unstable"]
 targets = ["x86_64-unknown-linux-gnu"]
-features = ["serde", "arbitrary", "slog", "v1", "v3", "v4", "v5", "v6", "v7", "v8"]
+features = ["serde", "arbitrary", "slog", "borsh", "v1", "v3", "v4", "v5", "v6", "v7", "v8"]
 
 [package.metadata.playground]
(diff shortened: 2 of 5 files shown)


KodrAus:
Thanks for working on this @grovesNL!

This is something we’d be happy to support in `uuid`, but since `borsch` is unstable we need to do a little extra work. If you look at how the `zerocopy` feature works we use a `RUSTFLAGS` config to toggle unstable features on. That makes them unsuitable for use in libraries but ok for applications. We’d need to so something similar here until there’s a stable API for `borsch`.

grovesNL:
Thanks @KodrAus! That sounds good to me, and the unstable Rust flags would be ok for my use case for now.

I updated the PR to require `uuid_unstable` for `borsh` (like `zerocopy`).

I guess another possibility is to keep the derives in my own crate if there was a way to pass a local type to `Uuid`, e.g. if `Uuid` was actually `Uuid<T = ()>(Bytes, PhantomData<T>)` or something to avoid the orphan rule. So we might have something like
```rust
struct CustomId;
type CustomUuid = Uuid<CustomId>;
impl BorshSerialize for CustomUuid { /* ... */ }
```

KodrAus:
Thanks for your patience on this one @grovesNL.

This looks good to me! The idea of using a "pivot type" like `Uuid<T>` is an interesting one I've used in internal code a few times, but probably a bit too much cognitive overhead for `uuid`.

PR #690, opened 2023-06-22
Fixing issue with Cloudflare Workers and wasm32-unknown-unknown when using now()

### LINKED ISSUE #689: 1.3.4 and 1.3.3 failing in Cloudflare Worker env for different reasons
I'm really in over my head on this one but here is my understanding.

Due to changes mentioned in #687, 1.3.4 does not work correctly in wasm32-unknown-unknown:

```
panicked at 'time not implemented on this platform', library/std/src/sys/wasm/../unsupported/time.rs:31:9
```

However, 1.3.3 also fails on Cloudflare Workers:

```
LinkError: WebAssembly.Instance(): Import #9 module="./index_bg.js" function="__wbg_now_5eac63e226f04e45" error: function import requires a callable
```

With the changes made in PR #688, all wasm32-unknown-unknown tests work fine. A very similar issue is brought up in [this issue](https://github.com/rustwasm/gloo/pull/249). Applying the same fix keeps all tests happy and allows v7_now() to be used in Cloudflare Workers. Here is the fix:

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = Date)]
    fn now() -> f64;
}

let now = now();
```

becomes:

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = Date, catch)]
    fn now() -> Result<f64, JsValue>;
}

let now = now().unwrap_throw();
```

Not sure how to test for this and I don't really understand what is causing the issue. However, the updated code works and passes all tests.

kmusick:
Fixes #689

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -262,17 +262,17 @@ pub(crate) const fn decode_unix_timestamp_millis(uuid: &Uuid) -> u64 {
     millis
 }
 
-#[cfg(all(feature = "std", feature = "js", target = "wasm32-unknown-unknown"))]
+#[cfg(all(feature = "std", feature = "js", all(target_arch = "wasm32", target_vendor = "unknown", target_os = "unknown")))]
 fn now() -> (u64, u32) {
     use wasm_bindgen::prelude::*;
 
     #[wasm_bindgen]
     extern "C" {
-        #[wasm_bindgen(js_namespace = Date)]
-        fn now() -> f64;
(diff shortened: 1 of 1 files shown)


KodrAus (approved):
Thanks @kmusick!

I’m not entirely sure what’s going on here either yet. I think our treatment of wasm targets has been historically over-simplified.

I also noticed we have a `dbg!` statement in that function too which seems like it really shouldn’t be there so I’ll spend some time in there too and try unpick it all.

PR #693, opened 2023-07-12
Add #[inline] for Uuid::from_bytes[_ref] and Uuid::{as,into}_bytes

jrose-signal:
[As seen in this playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=caa6418cf6c80c8ccbf8c5521fccb5b4), neither `from_bytes` nor `from_bytes_ref` are currently inlined by the compiler even though they are `const`, resulting in less efficient generated code for these trivial functions, even with Thin LTO enabled. There's a slippery slope here for what counts as "trivial", but the functions that *do not change representations* are a pretty safe bet.

(I didn't include the Builder versions of the same functions, because the *next* thing you do with a Builder is likely to be a function call anyway, but it could go either way.)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -412,6 +412,7 @@ impl Uuid {
     /// # Ok(())
     /// # }
     /// ```
+    #[inline]
     pub const fn from_bytes(bytes: Bytes) -> Uuid {
         Uuid(bytes)
     }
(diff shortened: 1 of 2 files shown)


KodrAus (approved):
Thanks @jrose-signal! This looks good to me.

PR #694, opened 2023-07-12
Fix macro hygiene

teohhanhui:
Re-export Ok and Err

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -273,6 +273,11 @@ mod macros;
 #[cfg(feature = "macro-diagnostics")]
 pub extern crate uuid_macro_internal;
 
+#[doc(hidden)]
+pub mod __macro_support {
+    pub use crate::std::result::Result::{Err, Ok};
+}
+
 use crate::std::convert;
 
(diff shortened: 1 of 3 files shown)


teohhanhui:
Easy way to trigger the problem:

```rust
use anyhow::Ok;
use uuid::{uuid, Uuid};

const SOME_UUID: Uuid = uuid!("00000000-0000-0000-0000-000000000000");
```

Example of error message when running into this macro hygiene issue:

```
error[E0532]: expected tuple struct or tuple variant, found function `Ok`
```

teohhanhui:
ping @KodrAus

KodrAus:
Thanks for the PR @teohhanhui!

This looks good to me, but would you be happy to add a test case [here](https://github.com/uuid-rs/uuid/tree/main/tests/ui/compile_pass) that exercises it? You could replace the `anyhow::Ok` in your repro with `type Ok = ()` or something like that.

teohhanhui:
Done.

KodrAus (approved):
Thanks for fixing this up @teohhanhui!

PR #699, opened 2023-07-26
Added conversion for Vec<u8>

Jeremiah-Griffin:
Hello,

An `impl From<Uuid> for Vec<u8>` is quite useful for crates like Diesel which need such a conversion to work with its codegen of database models.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -11,10 +11,17 @@
 
 //! Adapters for alternative string formats.
 
+use alloc::vec::Vec;
 use crate::{
     std::{borrow::Borrow, fmt, ptr, str},
     Uuid, Variant,
 };
+
+impl From<Uuid> for Vec<u8>{
+    fn from(value: Uuid) -> Self {
+        value.as_bytes().to_vec()
+    }
+}
 
 impl std::fmt::Debug for Uuid {
     #[inline]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -272,6 +272,7 @@ mod macros;
 #[doc(hidden)]
 #[cfg(feature = "macro-diagnostics")]
 pub extern crate uuid_macro_internal;
+extern crate alloc;
 
 #[doc(hidden)]
 pub mod __macro_support {


src/fmt.rs
```diff
@@ -11,11 +11,18 @@
 
 //! Adapters for alternative string formats.
 
+use alloc::vec::Vec;
 use crate::{
     std::{borrow::Borrow, fmt, ptr, str},
     Uuid, Variant,
 };
 
+impl From<Uuid> for Vec<u8>{
```
KodrAus, line 20: This will need to be conditionally compiled, with a `#[cfg(feature = “std”])`.

src/lib.rs
```diff
@@ -272,6 +272,7 @@ mod macros;
 #[doc(hidden)]
 #[cfg(feature = "macro-diagnostics")]
 pub extern crate uuid_macro_internal;
+extern crate alloc;
```
KodrAus, line 275: This will also need to be conditionally compiled.

KodrAus (commented):
Thanks for the PR @Jeremiah-Griffin!

We’ll need to make this impl conditional so that no-std users don’t get it.

Right now we have a `std` feature that would cover it. If you’re keen though you could add an `alloc` feature that’s implied by `std` and just brings in the `alloc` crate.

src/fmt.rs
```diff
@@ -11,10 +11,27 @@
...
+use alloc::vec::Vec;
+use crate::{Bytes, std::{borrow::Borrow, fmt, ptr, str}, Uuid, Variant};
+
+impl From<Uuid> for Vec<u8>{
+    fn from(value: Uuid) -> Self {
+        value.as_bytes().to_vec()
+    }
+}
+
+impl From<Vec<u8>> for Uuid{
```
KodrAus, line 23: I don't think we should add this side of the conversion. Converting slices into `Uuid`s is fallible, but we can't express that through this trait.

Cargo.toml
```diff
@@ -70,6 +70,9 @@ fast-rng = ["rng", "rand"]
 sha1 = ["sha1_smol"]
 md5 = ["md-5"]
 
+[dependencies]
+diesel = {version = "2.1.0", features = ["sqlite"]}
```
KodrAus, line 74: Hmm, digging back into this it looks like `diesel` actually already depends on `uuid`, so I don't think we can add it as a dependency here. You might not actually need to do anything in this library to use it with `diesel` besides enable its `uuid` feature. It looks like their support is implemented here: https://github.com/diesel-rs/diesel/blob/a122e7b88a1bc1b427dde34ed36ce3f36fad772e/diesel/src/pg/types/uuid.rs#L13.

KodrAus:
Thanks again for the PR @Jeremiah-Griffin. I've ended up merging #703 which adds some `Vec` conversions. We won't be able to release this patch as-is, but if you're using your branch in a project of yours you should be able to continue doing so even though the PR is closed.

PR #700, opened 2023-08-05
Add impl From<Uuid> for String under the std feature flag

brahms116:
This PR adds the implementation of `From<Uuid> for String` under the "std" feature flag. The motivation for this is that it allows Uuids to be easily converted to a string when using crates like sqlx where it utilize the `Into` and `From` traits for conversions.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -16,6 +16,9 @@ use crate::{
     Uuid, Variant,
 };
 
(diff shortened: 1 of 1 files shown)


KodrAus (approved):
Thanks for the PR @brahms116!

Since we already inherit a `ToString` method, this seems reasonable to me.

PR #705, opened 2023-08-11
make ClockSequence wrap correctly

### LINKED ISSUE #702: Wrapping behavior of `impl ClockSequence for Context`
Hi there!  First of all, sorry for not using the bug report template; the link in the contribution guide seems broken.

Anyway, I was just browsing your code and noticed a minor oddity in the implementation of `ClockSequence` for `Context`, in [timestamp.rs:441](https://github.com/uuid-rs/uuid/blob/97b7f07b3ea7455409e44f0bb4bc697009b7b844/src/timestamp.rs#L441).  Based on what the comment there states, `generate_sequence()` is supposed to just truncate the counter to 14 bits before returning it.  However, it actually takes the modulus with `0x1fff` instead of `0x2000`, which decreases entropy by one and compiles to a division instead of a simple bitwise AND.

Is this intentional?  I realize I'm being a bit pedantic here, but it can't hurt to point out.

fef1312:
As pointed out in the original issue, the implementation of `ClockSequence` for `Context` incorrectly wrapped after `0x3ffe` instead of `0x3fff`.  Fixes #702.

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -438,7 +438,7 @@ pub mod context {
             // increment the clock sequence we want to wrap once it becomes larger
             // than what we can represent in a "u14". Otherwise there'd be patches
             // where the clock sequence doesn't change regardless of the timestamp
-            self.count.fetch_add(1, Ordering::AcqRel) % (u16::MAX >> 2)
+            self.count.fetch_add(1, Ordering::AcqRel) & (u16::MAX >> 2)
         }
     }
 }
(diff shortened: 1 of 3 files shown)


KodrAus (approved):
Thanks @fef1312!

KodrAus:
It looks like we've got some tests for this function afterall. This one is now failing:

```
v1::tests::test_new_context
```

fef1312:
Ah, I missed that one because I forgot to enable the feature while testing.  Sorry!  It's just a matter of changing the counter's initializer to the correct value though, I'll fix it soon.

KodrAus:
Thanks for fixing this up @fef1312! I should probably move those tests to live with the type itself. I was surprised not to see any for it, and didn't realise that they were just way off in a different module.

Issue #710, opened 2023-10-15
Add bytemuck support


John-Toohey:
I am using `Uuid`s with `bevy_ggrs_rollback`, which requires `bytemuck` compatible types for  serialization and deserialization. By adding the `bytemuck` feature, it would thus become possible to use `Uuid` with said library (and other similar libraries). Whilst it is possible to do this by using a newtype or the `as_u128` and `from_u128` methods directly, I believe that adding bytemuck support will make code using bytemuck cleaner.

Thank you for your time,

John T

KodrAus:
This is available as of `1.5.0`

PR #711, opened 2023-10-15
Added `bytemuck` support

John-Toohey:
This pull request adds support for the `bytemuck` crate, hence fixing my earlier feature request, issue #710.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Continuous integration
 
 env:
   VERSION_FEATURES: "v1 v3 v4 v5 v6 v7 v8"
-  DEP_FEATURES: "slog serde arbitrary borsh zerocopy"
+  DEP_FEATURES: "slog serde arbitrary borsh zerocopy bytemuck"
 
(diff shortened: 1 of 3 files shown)


KodrAus:
Thanks for the PR @John-Toohey! I think our big-endian tests in CI are going to fail here for reasons unrelated to your PR. I'll fix those up separately.

Would it be possible to add `bytemuck` to the list of dependencies we check [here](https://github.com/uuid-rs/uuid/blob/main/.github/workflows/ci.yml#L5) too? Otherwise this looks good to me!

KodrAus:
I've merged #712 which should get CI back to green here. I think all we need to do is add `bytemuck` to our dependency list [here](https://github.com/uuid-rs/uuid/blob/main/.github/workflows/ci.yml#L5) and this should be good to go.

John-Toohey:
I have added bytemuck to the CI file; and have added `bytemuck` to the lib.rs documentation for easy discoverability

KodrAus (approved):
Thanks @John-Toohey! This looks good to me.

PR #716, opened 2023-10-30
feat: add `prost` support

### LINKED ISSUE #715: Add `prost::Message` implementation for `Uuid`
## Motivation
[Prost](https://github.com/tokio-rs/prost) is a [Protocol Buffers](https://developers.google.com/protocol-buffers/) implementation for the [Rust Language](https://www.rust-lang.org/). Protocol Buffers are used as the message exchange format for [gRPC](https://grpc.io/). Clients and servers need to build and parse protocol messages. The Prost crate has a code generator to transform .proto files into Rust types that implement the `prost::Message` trait to simplify message encoding/decoding. When working with `Uuid`s, it's common to have to write boilerplate to handle transforming between whatever the code gen'd type is and the proper `Uuid` type.

## Solution
The Prost crate has support to use existing Rust types rather than generating new types provided those existing types implement the `prost::Message` trait. Thus, the solution is to add a `prost::Message` implementation behind a feature flag.

```rust
// Raw encoding
#[cfg(feature = "prost-bytes")]
impl prost::Message for Uuid ..

// String Encoding
#[cfg(feature = "prost-hyphenated")]
impl prost::Message for Uuid ..
```

One challenge is that Protobuf users may require a particular encoding format, and so a decision would need to be made about which formats to support or whether to support all the encoding formats implemented by this crate. I have included the two more common formats in the above example, but perhaps a user would need an alternative encoding like urn.

## Alternatives
- Continue writing boilerplate to convert between a generated Uuid struct (typically the equivalent of `struct Uuid(Vec<u8>)` and the uuid-rs `Uuid` type.
- Implement a 3rd crate to get around the orphan trait rules; this still requires a newtype and some boilerplate but is a bit less boilerplate than the first alternative.

## Is it blocking?
No

## Anything else?
Prost is a rather mature crate in the tokio organization that has quite a bit of adoption and has likely cemented its place in the Rust ecosytem. Adding a feature to interoperate with prost is unlikely to result in feature churn.

aymanalqadhi:
Pull request to add support for `prost` crate by implementing `prost::Message` for `uuid::Uuid`.

Resolves #715

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,20 @@ rust-version = "1.57.0"
 rustc-args = ["--cfg", "uuid_unstable"]
 rustdoc-args = ["--cfg", "uuid_unstable"]
 targets = ["x86_64-unknown-linux-gnu"]
-features = ["serde", "arbitrary", "slog", "borsh", "v1", "v3", "v4", "v5", "v6", "v7", "v8"]
+features = [
+    "serde",
+    "arbitrary",
+    "slog",
+    "borsh",
+    "prost",
+    "v1",
+    "v3",
+    "v4",
+    "v5",
+    "v6",
+    "v7",
+    "v8",
+]
 
 [package.metadata.playground]
 features = ["serde", "v1", "v3", "v4", "v5", "v6", "v7", "v8"]
@@ -110,6 +123,10 @@ version = "0.6"
 [dependencies.borsh]
 optional = true
 version = "0.10.3"
+
+[dependencies.prost]
+optional = true
+version = "0.12"
 
 # Private
 # Don't depend on this optional feature directly: it may change at any time
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ homepage = "https://github.com/uuid-rs/uuid"
 name = "uuid"
 readme = "README.md"
 repository = "https://github.com/uuid-rs/uuid"
-version = "1.5.0" # remember to update html_root_url in lib.rs
+version = "1.5.1" # remember to update html_root_url in lib.rs
 rust-version = "1.57.0"
 
 [package.metadata.docs.rs]
@@ -39,7 +39,8 @@ features = [
     "arbitrary",
     "slog",
     "borsh",
-    "prost",
+    "prost-string",
+    "prost-bytes",
     "v1",
     "v3",
     "v4",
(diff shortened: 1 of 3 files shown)


jwilm:
Hi @aymanalqadhi, 

Wow, I was not expecting someone else to make a PR for my issue! This is exciting to see.

I believe this only _partially_ addresses #715 because this only supports the _hyphenated string_ serialization. In my proposal, I suggested that we start with two feature flags, one for the hyphenated encoding and one for a binary encoding. I would be in support of merging a PR like this if the feature flag were renamed to something like `prost-hyphenated` to allow for additional encoding variants.

aymanalqadhi:
I think a better name for the feature would be `prost-string` or something, since basic hex-encoded values are also supported.

lilymara-onesignal:
Hi, can I ask why this PR got closed? It's a pretty nice feature to have for people using UUIDs in gRPC services, it would be great if we could get it merged in.

jwilm:
> Hi, can I ask why this PR got closed? It's a pretty nice feature to have for people using UUIDs in gRPC services, it would be great if we could get it merged in.

It was identified in the feature request issue that the approach of using feature flags would make serialization strategies mutually exclusive. This would prevent the ability to write applications that talk to services using alternative serialization strategies.

I think the next step would be considering how prost might be able to support such a use case without requiring newtypes.

aymanalqadhi:
What if a specific format (let's say binary) is used as the default, and is enabled automatically when specifying the `prost` feature. And a second feature (e.g., `prost-string`) can be used to switch to the string format?

KodrAus:
Crate features just aren't suitable for this kind of configuration because you don't have control over what features are activated in your dependency graph, so if you end up in a situation where two dependencies set incompatible features then you'll have no way to remedy things. Worse still, you won't know they're wrong until you've possibly already encoded protobufs with the wrong format, which is spongy enough on the wire that it may even just silently work.

If we can come up with an approach that isn't reliant on crate features then I'd be more than happy to revisit this 🙂

-----

For what it's worth, I tend to separate these kinds of interchange types from whatever richer model is used within a service itself and map between them. That way your protos can work in terms of the simplest possible datatype, and your service can work with the most useful. Rust makes that pattern both cheap and difficult to get wrong.

aymanalqadhi:
I totally get your point. The approach I suggested does not depend on mutually exclusive feature flags.

If the idea of using multiple feature flags is out of the question, then perhaps supporting a single format (string is probably the most used, albeit less efficient) should suffice for most users.

jwilm:
> If the idea of using multiple feature flags is out of the question, then perhaps supporting a single format (string is probably the most used, albeit less efficient) should suffice for most users.

As convenient as this would be _today_, I think it's not the right solution. Uuid is already 1.0 and values stability. Adding a prost flag feature that in any way makes assumptions about encoding format is eventually going to require a breaking change when a more robust solution becomes available.

PR #724, opened 2023-12-19
Upgrade borsh unstable dependency to v1.0 and make it stable

bgeron:
I checked manually that even with borsh =v1.0.0 the tests pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,10 +103,7 @@ version = "1.1.3"
 optional = true
 version = "0.6"
 
-# Public (unstable): Used in trait impls on `Uuid`
-# Unstable: also need RUSTFLAGS="--cfg uuid_unstable" to work
-# This feature may break between releases, or be removed entirely before
-# stabilization.
+# Public: Used in trait impls on `Uuid`
 [dependencies.borsh]
 optional = true
 version = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,8 @@ fast-rng = ["rng", "rand"]
 sha1 = ["sha1_smol"]
 md5 = ["md-5"]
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -441,7 +441,7 @@ pub enum Variant {
 )]
 #[cfg_attr(
     all(uuid_unstable, feature = "borsh"),
```
KodrAus, line 443:
```suggestion
    feature = "borsh",
```

The `uuid_unstable` gate should no longer be necessary now that `borsh` is stable.

KodrAus (commented):
Thanks @bgeron! It looks like `borsh` has been stable for a little while now, but still a huge props to the authors for pulling it all together 👏

bgeron:
> Thanks @bgeron! It looks like `borsh` has been stable for a little while now, but still a huge props to the authors for pulling it all together 👏

Very much agree! :clap: I'm not involved in borsh but I'm thinking to start using it.

- I changed more to make feature `borsh` stable, per your suggestion.

- I noticed that it wasn't possible to compile `uuid` with only the `borsh` feature because it didn't yet imply `dep:borsh-derive`, so I added that implication. Now a dummy program that depends on both uuid and borsh works. I'm not so experienced with Cargo feature implication but it seems to work.

Cargo.toml
```diff
@@ -71,6 +71,8 @@ fast-rng = ["rng", "rand"]
 sha1 = ["sha1_smol"]
 md5 = ["md-5"]
 
+borsh = ["dep:borsh", "dep:borsh-derive"]
```
KodrAus, line 74: The fact that this works on our MSRV now is great :+1: There's so much I can clean up elsewhere with our dependencies.

KodrAus (approved):
Thanks @bgeron! This looks good to me.

Issue #638, opened 2022-10-09
Tracking issue for defmt


KodrAus:
This issue is a place to keep anybody using the unstable `defmt` feature up-to-date on any changes being made to it.

In order to use `defmt`, you need to opt-in not just to the feature, but to the `uuid_unstable` RUSTFLAGS configuration too. We can make breaking changes to `defmt` at any point, but still want to be considerate of anybody making use of it, so will try keep open about any version bumps or other breakage that's coming.

KodrAus:
This wasn't implemented in the end, because `defmt` can't be depended on by libraries.

guissalustiano:
Sorry @KodrAus, can you explain a little further why `defmt` can't be depended on by libraries?

KodrAus:
@guissalustiano It looks like I never fully linked up or articulated what happened when we tried to do this last time. The problem is described here: https://github.com/knurling-rs/defmt/issues/673#issuecomment-1142150012

As I understand it, `defmt` is only meant for microcontrollers and doesn't build on full OSes. That makes it unsuitable as a dependency for libraries targeting broad compatibility. Even as an optional dependency it breaks `--all-features`, which I do want to keep working.

PR #732, opened 2024-02-03
[feat]: add (optional) defmt support

guissalustiano:
The [defmt](https://github.com/knurling-rs/defmt) is an important crate in the embedded world, it implements a highly efficient logging encoding using deferred formatting and string compression. But for that, they need a special format implementation.

This PR implements the `defmt::Format` for `Uuid` based on feature configuration.

Example output after run `probe run`
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,8 @@ atomic = ["dep:atomic"]
 
 borsh = ["dep:borsh", "dep:borsh-derive"]
 
+defmt = ["dep:defmt"]
+
 # Public: Used in trait impls on `Uuid`
 [dependencies.bytemuck]
 version = "1.14.0"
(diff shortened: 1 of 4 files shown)


guissalustiano:
I don't know why in my first search I didn't find anything. But this already was discussed [here](https://github.com/uuid-rs/uuid/issues/638)

guissalustiano:
> As I understand it, defmt is only meant for microcontrollers and doesn't build on full OSes. That makes it unsuitable as a dependency for libraries targeting broad compatibility. Even as an optional dependency it breaks --all-features, which I do want to keep working.
> [@KodrAus  at #638 ](https://github.com/uuid-rs/uuid/issues/638#issuecomment-1925972796)

PR #734, opened 2024-02-04
feat: avoid copying in TryFrom<Vec<u8>>

### LINKED ISSUE #733: TryFrom<Vec<u8>> makes an unnecessary copy
As `Vec<u8>` can be safely converted to `Bytes` instead of copying its contents.

pjenvey:
Closes #733

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -281,6 +281,9 @@ pub mod __macro_support {
 
 use crate::std::convert;
 
+#[cfg(feature = "std")]
+use crate::{error::ErrorKind, std::convert::TryInto};
+
 pub use crate::{builder::Builder, error::Error};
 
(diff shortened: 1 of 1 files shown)


KodrAus:
Hi @pjenvey :wave:

For what it's worth it looks like these generate roughly equivalent code, but the existing implementation is slightly better: https://godbolt.org/z/138rGesMz

We end up creating a copy of the bytes anyway since `uuid` takes ownership of them. Did you observe some surprising behavior attempting to use these `TryFrom` conversions?

KodrAus:
Thanks for the PR @pjenvey. I couldn't find a compelling reason to change this; they seemed to generate the same code, so will close this one to keep the source a bit simpler.

If you do notice any real-world impact here please feel free to re-open.

PR #735, opened 2024-02-04
serde serialize_with support

dakaizou:
**I'm submitting a** feature

Description
========
Adds the ability to serialize a `Uuid` as a `Simple`/`Urn`/`Braced`. `Uuid::parse_str` deserialize them correctly, but `Uuid` is always serialized as `Hyphenated`. Users would be able to decide which format to serialize to after this PR.

Example:
 ```rust
 #[derive(serde_derive::Serialize)]
 struct Struct {
     // This will be serialize as uuid::fmt::Simple
     #[serde(serialize_with = "uuid::serde::simple")]
     id: uuid::Uuid,
}
 ```

Tests
========
3 new tests for the feature were added.

crate::external::serde_support::serde_tests::test_serialize_as_simple
crate::external::serde_support::serde_tests::test_serialize_as_braced
crate::external::serde_support::serde_tests::test_serialize_as_urn

Related Issue(s)
========
- https://github.com/serde-rs/serde/issues/723
  Before this issue is resolved, types like `Option<Uuid>` require their own functions to be serialized correctly.

### DIFF
--- a/src/external/serde_support.rs
+++ b/src/external/serde_support.rs
@@ -207,6 +207,72 @@ pub mod compact {
     }
 }
 
+/// Serialize from a [`Uuid`] as a `uuid::fmt::Simple`
+///
+/// [`Uuid`]: ../../struct.Uuid.html
+///
+/// # Example
+///
+/// ```rust
+/// #[derive(serde_derive::Serialize)]
+/// struct Struct {
+///     // This will be serialize as uuid::fmt::Simple
+///     #[serde(serialize_with = "uuid::serde::simple")]
+///     id: uuid::Uuid,
+/// }
+///
+/// ```
+pub fn simple<S>(u: &crate::Uuid, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    serde::Serialize::serialize(u.as_simple(), serializer)
+}
+
+/// Serialize from a [`Uuid`] as a `uuid::fmt::Urn`
+///
+/// [`Uuid`]: ../../struct.Uuid.html
+///
+/// # Example
+///
+/// ```rust
+/// #[derive(serde_derive::Serialize)]
+/// struct Struct {
+///     // This will be serialize as uuid::fmt::Urn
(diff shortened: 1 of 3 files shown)


KodrAus (approved):
Nice! This looks good to me @dakaizou.

I don't think we need to do `deserialize_with` support, since the parser will very quickly identify the appropriate format to use. We can also add deserialize support in a non-breaking way in the future if we want to.

KodrAus (changes requested):
Actually, having another look at the existing `compact` module, I think we should follow the same pattern here. Instead of top-level functions in `uuid::serde`, let's use modules with a `serialize` and `deserialize` function in them.

dakaizou:
> ... I don't think we need to do `deserialize_with` support, since the parser will very quickly identify the appropriate format to use. ...

This was exactly what I was thinking, since they can all be deserialized with `Uuid::deserialize`, I only created functions for serialization.


> I think we should follow the same pattern here. Instead of top-level functions in uuid::serde, let's use modules with a serialize and deserialize function in them.

`compact::deserialize` only deserializes `[u8; 16]`, so it makes sense to put them in a mod for `#[serde(with="compact")]`. 
For `Simple`/`Braced`/`Urn`, do you think it's better to follow the same pattern? (e.g. rejecting the `Braced` format from `Simple`'s deserialize)

KodrAus:
> For Simple/Braced/Urn, do you think it's better to follow the same pattern? (e.g. rejecting the Braced format from Simple's deserialize)

It looks like we don't have exposed methods for specific parsers, but we could make them `pub(crate)` and use those for these methods. I think it'll be more performant and align better with expectations if we only parse exactly the format requested.

dakaizou:
- create two more methods in `parser`: `parse_braced`, `parsed_urn`
- fix the error returned by `visit_seq` in `Uuid::deserialize`
- create `simple`, `braced`, `urn` mod in `serde_support`

KodrAus (approved):
Thanks for putting this together @dakaizou! This is some great work 🙂

PR #736, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


KodrAus:
Thanks for the PR @oriontvv. I haven't traditionally found Dependabot very useful, particularly for libraries. It generates a lot of noise for transitive dependencies and rather than merge any of its PRs I'd rather do updates myself.

If it's possible to configure it to only open a single PR for minor releases of direct dependencies then I think it would be much more useful.

oriontvv:
We can limit number of opened PR to 1 (10 in the request). We also could skip patch updates as well [update-types: major](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

kinggoesgaming:
> We can limit number of opened PR to 1 (10 in the request). We also could skip patch updates as well [update-types: major](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

I do have an issue the the `major` only policy as it can unintentionally introduce breaking changes.

KodrAus:
There are some dependencies we can take new major versions of, although would expect simply updating them in our manifest to result in broken builds. I think a good starting point here could be running `cargo outdated` in a separate CI job we can inspect. It would be neat if we had a UX that surfaced dependency updates as some sort of status indicator rather than pull requests so we can act accordingly.

KodrAus:
I've added `cargo outdated` to a separate workflow that's flagged some dependencies we can look at updating: https://github.com/uuid-rs/uuid/actions/runs/8352241415/job/22862011793#step:4:5

I think that will suit our purposes, at least for now, so will go ahead and close this one as something we won't add.

Thanks again for the PR @oriontvv!

PR #739, opened 2024-02-16
Add bitcode support

tbillington:
This pr adds support for [`bitcode`](https://github.com/SoftbearStudios/bitcode), commonly used by bevy games for it's more compact representation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Continuous integration
 
 env:
   VERSION_FEATURES: "v1 v3 v4 v5 v6 v7 v8"
-  DEP_FEATURES: "slog serde arbitrary borsh zerocopy bytemuck"
+  DEP_FEATURES: "slog serde arbitrary borsh zerocopy bytemuck bitcode"
 
 on:
   pull_request:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,6 +99,11 @@ version = "2"
 optional = true
 version = "1.1.3"
 
+# Public: Used in trait impls on `Uuid`
+[dependencies.bitcode]
+version = "0.5.1"
+optional = true
+
 # Public (unstable): Used in `zerocopy` derive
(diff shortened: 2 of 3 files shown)


KodrAus:
Thanks for the PR @tbillington.

Since the `bitcode` library is unstable we can't add support to `uuid` without also requiring the `uuid_unstable` rustflag be set too. The `zerocopy` support has an example of how to do this. Needing to set a flag at build-time makes these unstable features largely unsuitable for libraries, but can be used by applications that are willing to opt-in to it.

Would you be happy to add the extra `uuid_unstable` cfg to this?

tbillington:
Oh interesting, I can look into that 👍 

When you say the library in unstable, what exactly do you mean?

KodrAus:
Ah I just mean that the library is pre-1.0, so breaking changes are still expected. If the library does make breaking changes then we'll either need to make a breaking change in `uuid` too to adopt it, or create another feature to support it. Having the unstable opt-in RUSTFLAG is a balance between adding support for new libraries, and insulating `uuid` from breaking changes in its dependencies.

caibear:
Another solution is to specify `bitcode = { version = "0", features = ["derive"], default-features = false, optional = true }` and only use `#[derive(Encode, Decode)]`. These apis are considered a stable subset of bitcode. Then users of uuid can choose the major version of bitcode.

tbillington:
Happy to make the change caibear suggested, would that still require `uuid_unstable`?

KodrAus:
Hmm, I wouldn't be comfortable taking on this kind of stability policy in `uuid`. I'd be more than happy to accept this under the same unstable policy as `zerocopy` though.

vnghia:
Hello I've created a PR to add support for `UUID` in `bitcode`. I think it is better if we add the support there instead of here. https://github.com/SoftbearStudios/bitcode/pull/34

KodrAus:
Just coming back through some triage. This seems like it would be better supported outside of `uuid`, or if it is in `uuid`, as an unstable feature. I'll go ahead and close this one now, but we can revisit in the future.

Issue #717, opened 2023-11-05
v7 should have a counter option?


rogusdev:
Per the new draft "With this method rand_a section of UUIDv7 SHOULD be utilized as fixed-length dedicated counter bits that are incremented by one for every UUID generation."

That's in the "Fixed-Length Dedicated Counter Bits (Method 1)" section under https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-04#monotonicity_counters which is directly linked from https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-04#section-5.2 under the rand_a explanation.

To my reading, that means it should have a counter, like v6 (and v1), but instead the current v7 builder implementation https://docs.rs/uuid/latest/src/uuid/v7.rs.html#47-53 is just random for rand_a. Am I missing something on this one? If not, and this is a desirable feature, I might look into putting up a PR to add the counter support.

KodrAus:
Hi @rogusdev :wave:

I'd be keen to go catch up on some of the discussion around that. Those sections look like recommendations rather than requirements, so I think `uuid`'s current implementation that fills the entire 74 bits of `rand_a` and `rand_b` with random data is still compliant.

It seems a bit overkill to me to need a clock, a source of randomness, _and_ a monotonic counter to generate V7 UUIDs, but we could consider adding `*_monotonic` variants of the `v7` methods that accept a `ClockSequence` like `v1` and `v6` do.

KodrAus:
The test vector for UUIDv7 appears to use a fully random value for `rand_a` and `rand_b`: https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-04#name-example-of-a-uuidv7-value

I think we should consider supporting this, but don't think we need to block stabilizing the current APIs on it.

rogusdev:
From my perspective, while the main point of uuid v7 is to put the clock at the front, so that db indexes can btree on them properly, but counters add additional safety mechanisms that are quite to my taste: if you are generating ids in a single thread, you are *guaranteed* that no duplicates can happen, if there are less than the counter limit per millisecond. While the degree of randomness involved otherwise is certainly extremely unlikely to have duplicates, I am a big fan of guarantees.

rogusdev:
That said, "SHOULD" is indeed a recommendation, rather than requirement, and they list multiple alternatives. So I support having a separate set of ctor functions to call for the counter version(s).

[...]

PR #755, opened 2024-05-13
Support counters in v7 UUIDs

KodrAus:
For #717

This PR sketches out a solution to make the v7 support respect the input counter, while retaining existing semantics around filling with random data. It requires a bit of fudgery to keep compatibility with existing APIs, but I think the overall solution works out ok. I need to actually fill it all in fully and then explore ways to better integrate the existing `Timestamp` API, which may involve deprecating bits.

The gist of it is to add a `usable_bits` method to `ClockSequence`, so it can tell how many bits a given counter occupies. The UUID constructor can then do appropriate shifts and fill the rest with random bytes. That way a caller creating a v7 UUID can decide how many bits to dedicate to sequencing and how many to randomness.

As part of this PR I'm expecting to find some solution to re-seeding sequences, which will probably remain an implementation detail of the given `ClockSequence`.

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -117,8 +117,12 @@ impl Timestamp {
     }
 
     /// Construct a `Timestamp` from a Unix timestamp and up to a 16-bit counter, as used in version 7 UUIDs.
-    pub fn from_unix(context: impl ClockSequence<Output = u16>, seconds: u64, nanos: u32) -> Self {
-        Self::from_unix_128(context, seconds, nanos)
+    pub fn from_unix(
+        context: impl ClockSequence<Output = u16>,
+        seconds: u64,
+        subsec_nanos: u32,
+    ) -> Self {
+        Self::from_unix_128(context, seconds, subsec_nanos)
     }
 
     /// Construct a `Timestamp` from a Unix timestamp and up to a 128-bit counter, as used in version 7 UUIDs.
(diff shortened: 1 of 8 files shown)


src/timestamp.rs
```diff
@@ -434,6 +478,10 @@ pub mod context {
             // where the clock sequence doesn't change regardless of the timestamp
             self.count.fetch_add(1, Ordering::AcqRel) & (u16::MAX >> 2)
         }
+
+        fn usable_bits(&self) -> usize {
+            14
```
KodrAus, line 483: We need to add another implementation here with a 42bit counter that's intended to be used by v7 UUIDs.

src/v7.rs
```diff
@@ -41,15 +43,41 @@ impl Uuid {
     /// );
     /// ```
     ///
+    /// A v7 UUID can also be created with a counter to ensure batches of
+    /// UUIDs created together remain sortable:
+    ///
+    /// ```rust
+    /// # use uuid::{Uuid, Timestamp, Context};
+    /// let context = Context::new(42);
+    /// let uuid1 = Uuid::new_v7(Timestamp::from_unix(&context, 1497624119, 1234));
```
KodrAus, line 52: These examples should use `Timestamp::from_unix_wide` with a 42bit counter to steer users wanting counters in the right direction.

KodrAus:
I've got a counter implemented, but looks like performance here could use some improvement compared to [`uuid7`](https://github.com/LiosK/uuid7-rs):

```
running 2 tests
test new_v4   ... bench:          24.69 ns/iter (+/- 0.43)
test uuid7_v4 ... bench:          14.31 ns/iter (+/- 0.06)

test now_v7   ... bench:          56.86 ns/iter (+/- 1.04)
test uuid7_v7 ... bench:          38.03 ns/iter (+/- 0.59)
```

KodrAus:
Ok, I think that's better now:

```
test new_v4   ... bench:           8.35 ns/iter (+/- 0.34)
test uuid7_v4 ... bench:          14.38 ns/iter (+/- 0.78)

test now_v7   ... bench:          36.51 ns/iter (+/- 2.60)
test uuid7_v7 ... bench:          38.31 ns/iter (+/- 2.35)
```

KodrAus:
I need to add some tests for the new v7 context, and make sure we're shifting it around the variant field correctly.

PR #764, opened 2024-06-29
Use const identifier in uuid macro

### LINKED ISSUE #756: uuid!() macro can't handle a constant &str passed in
```rust
const TEST: &str = "00000000-0000-0000-0000-000000000000";
let uid: Uuid = uuid::uuid!(TEST);
```

no rules expected the token `TEST`
no rules expected this token in macro call
macros.rs(16, 14): while trying to match meta-variable `$uuid:literal`

Vrajs16:
The code below works when the user doesn't enable `macro-diagnostics` feature.

It's not possible to convert the value of identifier into a string literal at compile time, but is there another way we can go about this so we can give more details when this feature is enabled. 

If not, we can just paste the same code inside the macro where macro-diagnostics is enabled so that at least there wouldn't be a compile time error.
```rust
($uuid:ident) => {{
    const OUTPUT: $crate::Uuid = match $crate::Uuid::try_parse($uuid) {
        $crate::__macro_support::Ok(u) => u,
        $crate::__macro_support::Err(_) => panic!("invalid UUID"),
    };
    OUTPUT
}};
```

Closes #756

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -4,7 +4,7 @@ macro_rules! define_uuid_macro {
         #[cfg(feature = "macro-diagnostics")]
         #[macro_export]
         macro_rules! uuid {
-            ($uuid:ident) => {{
+            ($uuid:expr) => {{
                 const OUTPUT: $crate::Uuid = match $crate::Uuid::try_parse($uuid) {
                     $crate::__macro_support::Ok(u) => u,
                     $crate::__macro_support::Err(_) => panic!("invalid UUID"),
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -20,6 +20,13 @@ macro_rules! define_uuid_macro {
                 };
                 OUTPUT
             }};
+            ($uuid:ident) => {{
+                const OUTPUT: $crate::Uuid = match $crate::Uuid::try_parse($uuid) {
+                    $crate::__macro_support::Ok(u) => u,
+                    $crate::__macro_support::Err(_) => panic!("invalid UUID"),
+                };
+                OUTPUT
+            }};
         }
     }
 }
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks @Vrajs16! We should be able to support this in the procedural macro too by modifying this block here: https://github.com/uuid-rs/uuid/blob/main/macros/src/lib.rs#L70-L83.

We'd change it to parse a [`syn::Expr`](https://docs.rs/syn/latest/syn/enum.Expr.html) instead of a `syn::Lit`. If we match a `syn::ExprLit { lit: syn::Lit::Str(str) }` from that expression then we'd follow the old path that parses the string, otherwise we generate the same code as what you're doing in the `macro_rules` macro.

Is this something you'd like to take a shot at implementing? If not, I'll be happy to merge this change once it passes CI and update the macros myself.

Vrajs16:
Hi @KodrAus, not sure how I would generate the same the code in `build_uuid` function.
I was thinking, when a user puts a variable inside the `uuid!` macro, macro rules will match `($uuid:ident) => ...` and then we just use the same code as the below:
```rust
const OUTPUT: $crate::Uuid = match $crate::Uuid::try_parse($uuid) { 
    $crate::__macro_support::Ok(u) => u,
    $crate::__macro_support::Err(_) => panic!("invalid UUID"),
};
OUTPUT
```

KodrAus:
Ah, sorry I'd forgotten we still wrap the procedural macro in a regular macro-rules one. Duplicating the match arm in the case where `macro-diagnostics` is enabled should be all we need.

src/macros.rs
```diff
@@ -20,6 +20,13 @@ macro_rules! define_uuid_macro {
                 };
                 OUTPUT
             }};
+            ($uuid:ident) => {{
```
KodrAus, line 23:
```suggestion
            ($uuid:expr) => {{
```

We can generalize this to match any expression and it will still work.

src/macros.rs
```diff
@@ -4,6 +4,13 @@ macro_rules! define_uuid_macro {
         #[cfg(feature = "macro-diagnostics")]
         #[macro_export]
         macro_rules! uuid {
+            ($uuid:ident) => {{
```
KodrAus, line 7:
```suggestion
            ($uuid:expr) => {{
```

We can use expr here too.

KodrAus (approved):
Thanks @Vrajs16! This looks good to me.

PR #774, opened 2024-11-25
deps: replace rand with fastrand for fast-rng feature

roderickvd:
Replace the `rand` crate with `fastrand` in the `fast-rng` feature path. `fastrand` is a simpler and lighter-weight RNG implementation that better suits the needs of the `fast-rng` feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ v8 = []
 js = ["dep:wasm-bindgen", "getrandom?/js"]
 
 rng = ["dep:getrandom"]
-fast-rng = ["rng", "dep:rand"]
+fast-rng = ["rng", "dep:fastrand"]
 
 sha1 = ["dep:sha1_smol"]
 md5 = ["dep:md-5"]
@@ -131,9 +131,9 @@ optional = true
 version = "0.2"
 
 # Private
-[dependencies.rand]
+[dependencies.fastrand]
 optional = true
-version = "0.8"
+version = "2.2"
 
 # Private
 [dependencies.md-5]
--- a/src/rng.rs
+++ b/src/rng.rs
@@ -14,7 +14,7 @@ pub(crate) fn u128() -> u128 {
 
     #[cfg(feature = "fast-rng")]
     {
-        rand::random()
+        fastrand::u128()
     }
 }
 
@@ -34,7 +34,7 @@ pub(crate) fn u16() -> u16 {
 
     #[cfg(feature = "fast-rng")]
     {
-        rand::random()
+        fastrand::u16()
     }
 }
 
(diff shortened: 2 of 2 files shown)


KodrAus:
Thanks for the PR @roderickvd! We try stick to CSPRNG implementations in `uuid` for generating random data, so unfortunately I don't think we can use `fastrand`, which is not cryptographically secure. The `fast-rng` feature itself only lets us pull in more dependencies to support faster RNG, not weaken the algorithm we use.

roderickvd:
I understand that rationale. Maybe add it to the documentation for future onlookers like me?

I was thrown off because in some other crates, including `rand` itself with `SmallRng`, not being cryptographically secure is an acceptable compromise for a lighter-weight RNG. 

I know you were not using `SmallRng` before but I thought the feature-flag implied that desire.

roderickvd:
Or, would you entertain having this as a feature flag *additional* to the current ones for those who don't need the cryptographic security?

KodrAus:
> I understand that rationale. Maybe add it to the documentation for future onlookers like me?

That's a fair point, it should've occurred to me that it could be interpreted that way when I picked the name (and just used the feature name `rand`). I actually just wanted to use `rand` for randomness because `getrandom` is _very_ slow on Windows, but was met with complaints about additional dependencies. The `fast-rng` feature was a compromise I'd still like to undo at some point in the future.

I think we can't really do this using Cargo features unfortunately, because they're additive, so while your code may find a weaker RNG acceptable, other code that also uses UUIDs in your binary may not. I think for now, the best way to use a faster RNG is to use `uuid::Builder`, [like in this example](https://docs.rs/uuid/latest/uuid/struct.Builder.html#examples).

roderickvd:
Thanks for the hint on the builder. 

I guess what we *could* do is have a `fast-rng` _and_ `insecure-rng`. The one would pull in `rand` in addition to `getrandom` and the other `fastrand` in addition to `getrandom`. Both `rand` and `fastrand` depend on `getrandom` for seeding so that's OK. 

I could whip it up if you find the idea interesting. Otherwise don't worry about closing this.

KodrAus:
> I guess what we could do is have a fast-rng and insecure-rng.

I'm not immediately against this idea, but still a bit uncomfortable with it making a decision about the randomness of UUIDs on behalf of your entire dependency tree. I think we should be trying to move away from using Cargo features for what are effectively API configurations, and instead encourage using the `Builder` for plugging in alternative approaches.

KodrAus:
For the `Builder`, I think there isn't anything API-wise we need to add to `uuid`, you can already use https://docs.rs/uuid/latest/uuid/struct.Builder.html#method.from_random_bytes, where the source of those random input bytes can be from anywhere.

Thanks for putting this all together @roderickvd. I don't think we'll be supporting insecure RNGs at least through this approach, so will close this PR, but please feel free to open more with any other thoughts or ideas you had.

PR #777, opened 2024-12-22
Finish cut off docs

### LINKED ISSUE #776: Unfinished documentation
There's an unfinished sentence here: https://github.com/uuid-rs/uuid/blob/main/src/v7.rs#L34

"The data will be ..." I don't know what it will be :(

Thanks for a great and free product :)

KodrAus:
Closes #776


KodrAus:
Looks like our MSRV needs to be bumped to 1.63, and might need some wasm dependency updates too

Issue #770, opened 2024-09-30
NonZeroUuid


rick-de-water:
Is there a `NonZeroUuid` type, or has one been considered, similar to the other [`NonZero`](https://doc.rust-lang.org/std/num/struct.NonZero.html) types, such that the following is true?

```rust
assert_eq!(std::mem::size_of::<Option<NonZeroUuid>>(), std::mem::size_of::<Uuid>());
```

KodrAus:
Hi @rick-de-water 👋 A UUID with all zeroes is actually a valid value, it’s called the [nil UUID](https://www.rfc-editor.org/rfc/rfc9562.html#name-nil-uuid).

Are you looking at `Option<Uuid>` to avoid serializing them if they’re nil?

rick-de-water:
Of course, and an integer with a value of zero is also valid. There are however situation where you know that you will never use that value, and you can allow the compiler to do niche optimization if you inform it. This is what happens with the [`NonZero`](https://doc.rust-lang.org/std/num/struct.NonZero.html) types, and of course with references, since they also cannot be null.

Its actually quite common to have domain models that do not have NIL UUIDs, since they get generated at the moment of creation and never change. Having `NonZeroUuid`s (or `NonNilUuid`s if you want to be specific) can have a definite impact on the memory usage of a program.

KodrAus:
Coming back to this, it does seem like something that would be reasonable to add support for. I like the idea of calling it `NonNilUuid`. To avoid duplicating the entire API surface area though, I think it should be limited to conversions to and from `Uuid`s, and trait implementations that are also on `Uuid`.

PR #779, opened 2025-01-04
feat: Add `NonZeroUuid` type for optimized `Option<Uuid>` representation

ab22593k:
This commit introduces a new `NonZeroUuid` type that wraps a `Uuid` and guarantees that it is not the nil UUID. This allows for a more memory-efficient representation of `Option<Uuid>`, as `Option<NonZeroUuid>` takes up the same space as `Uuid`.

This change improves the memory efficiency of representing optional UUIDs and provides a more convenient way to work with non-nil UUIDs.

fixes: #770


src/external/nonzero_support.rs
```diff
@@ -0,0 +1,121 @@
...
+/// A UUID that is guaranteed not to be the nil UUID.
+///
+/// This is useful for representing optional UUIDs more efficiently, as `Option<NonZeroUuid>`
+/// takes up the same space as `Uuid`.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct NonZeroUuid(NonZeroU128);
+
+/// Error returned when attempting to create a `NonZeroUuid` from a nil UUID.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct NonZeroUuidError;
```
KodrAus, line 20: This should be a struct instead of a newtype, so that dependents of `uuid` can't construct it

src/external.rs
```diff
@@ -2,6 +2,8 @@
 pub(crate) mod arbitrary_support;
 #[cfg(feature = "borsh")]
 pub(crate) mod borsh_support;
+#[cfg(feature = "nonzero")]
+pub(crate) mod nonzero_support;
```
KodrAus, line 6: Since this isn't using an external library for support I think the module should live at the top-level.

KodrAus (commented):
Thanks for the PR @ab22593k! I've left a few comments to get started.

In addition to this, I think we should consider other trait implementations on this type:

- `Debug` and `Display` that are based on the underlying `Uuid`
- `serde::Serialize` and other external libraries that `uuid` optionally depends on

I'd also like to set the expectation that we won't be adding any methods to `NonNilUuid` besides construction from `Uuid`s and conversions to `Uuid`s, otherwise it'll end up duplicating `Uuid`s entire API on it.

It's also worth pointing out that the conversion will always succeed for any `Uuid` created via the `Uuid::new_v*` methods, since they set version/variant bits they're never nil.

src/external/nonzero_support.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+impl From<NonZeroUuid> for Uuid {
+    fn from(nz_uuid: NonZeroUuid) -> Self {
+        nz_uuid.get()
+    }
+}
+
+impl Deref for NonZeroUuid {
```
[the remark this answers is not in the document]
ab22593k: fixed

src/external/nonzero_support.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::ops::Deref;
+
+use crate::Uuid;
+
+/// A UUID that is guaranteed not to be the nil UUID.
+///
+/// This is useful for representing optional UUIDs more efficiently, as `Option<NonZeroUuid>`
+/// takes up the same space as `Uuid`.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct NonZeroUuid(NonZeroU128);
```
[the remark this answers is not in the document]
ab22593k: fixed

src/external.rs
ab22593k: fixed

Cargo.toml
```diff
@@ -79,6 +79,8 @@ atomic = ["dep:atomic"]
 
 borsh = ["dep:borsh", "dep:borsh-derive"]
 
+nonzero = []
```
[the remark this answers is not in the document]
ab22593k: fixed

src/external/nonzero_support.rs
ab22593k: Remove this for now, since used only on TryFrom trait

ab22593k:
PTL @KodrAus

KodrAus:
It looks like we've got some build failures with different feature combinations:

```text
Compiling proc-macro2 v1.0.93
     Compiling toml_edit v0.22.22
     Compiling quote v1.0.38
     Compiling syn v2.0.96
     Compiling toml v0.8.19
     Compiling serde_derive v1.0.217
     Compiling trybuild v1.0.101
     Compiling uuid v1.11.1 (/home/runner/work/uuid/uuid)
  error[E0599]: no function or associated item named `new_v4` found for struct `Uuid` in the current scope
     --> src/non_nil.rs:73:26
      |
  73  |         let uuid = Uuid::new_v4();
      |                          ^^^^^^ function or associated item not found in `Uuid`
      |
     ::: src/lib.rs:453:1
      |
  453 | pub struct Uuid(Bytes);
      | --------------- function or associated item `new_v4` not found for this struct
      |
  note: if you're trying to build a new `Uuid` consider using one of the following associated functions:
        builder::<impl Uuid>::nil
        builder::<impl Uuid>::max
        builder::<impl Uuid>::from_fields
        builder::<impl Uuid>::from_fields_le
        and 10 others
     --> src/builder.rs:72:5
      |
  72  |     pub const fn nil() -> Self {
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  ...
  98  |     pub const fn max() -> Self {
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  ...
  122 |     pub const fn from_fields(d1: u32, d2: u16, d3: u16, d4: &[u8; 8]) -> Uuid {
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ...
  168 |     pub const fn from_fields_le(d1: u32, d2: u16, d3: u16, d4: &[u8; 8]) -> Uuid {
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

ab22593k:
We should use `Uuid::builder()` since the `new_v4` method is not available.

src/non_nil.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// let uuid_again = Uuid::from(non_nil);
+    ///
+    /// assert_eq!(uuid, uuid_again);
+    /// ```
+    fn from(non_nil: NonNilUuid) -> Self {
+        Uuid::from_u128(non_nil.0.get())
+    }
+}
+
+impl From<Uuid> for NonNilUuid {
```
KodrAus, line 39: I don't think we should have a panicking `From` conversion and should instead use `TryFrom` like I think you had originally.

src/non_nil.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::{fmt, num::NonZeroU128};
+
+use crate::Uuid;
+
+/// A UUID that is guaranteed not to be the nil UUID.
+///
+/// This is useful for representing optional UUIDs more efficiently, as `Option<NonNilUuid>`
+/// takes up the same space as `Uuid`.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct NonNilUuid(NonZeroU128);
```
KodrAus, line 13: An associated `NonNilUuid::get()` that returned a `Uuid` would be helpful here I think.

Sorry, I think I might have lead you astray with an earlier comment about not having any methods instead of conversions. What I meant there was not having any methods like `new_v4`, `get_timestamp`, `get_version` etc.

KodrAus:
Thanks for working through this @ab22593k! I just had a few comments, but I think this is getting close now.

src/non_nil.rs
ab22593k: fixed

src/non_nil.rs
ab22593k: fixed, and thanks for your valuable comments.

KodrAus (approved):
Thanks @ab22593k! This looks good to me

PR #780, opened 2025-01-07
new from_u128 impl

Vrtgs:
use inherent methods of u128, rather than lots of bit shifting

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
-[target.wasm32-wasi]
+[target.wasm32-wasip1]
 runner = "wasmtime"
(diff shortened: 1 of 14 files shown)


KodrAus:
Thanks for the PR @Vrtgs! If you rebase on `main` that should fix up CI.

Vrtgs:
Done rebased! sorry it took so long, I didn't see any notifcation

KodrAus:
@Vrtgs would it be possible to cherry-pick just https://github.com/uuid-rs/uuid/pull/780/commits/da1b94d86eadee425a1cc2199ce1be272a126bc3 onto `main`? It's unfortunately quite difficult to review the real changes when they're buried in all the other changes there.

PR #788, opened 2025-01-19
use inherent to_be_bytes and to_le_bytes methods

Vrtgs:
use inherent to_be_bytes and to_le_bytes methods, rather than re-implementing them with lots of bitshifts


KodrAus (approved):
Thanks @Vrtgs!

PR #789, opened 2025-01-19
Fix links to namespaces in documentation

cstyles:
They're correct for [`Uuid::new_v3`](https://docs.rs/uuid/1.12.0/uuid/struct.Uuid.html#method.new_v3) but not for [`Uuid::new_v5`](https://docs.rs/uuid/1.12.0/uuid/struct.Uuid.html#method.new_v5).


KodrAus (approved):
Thanks @cstyles!

PR #797, opened 2025-02-05
Fix `wasm32` with `atomics`

### LINKED ISSUE #796: Failure to compile with `js` feature and `-C target-feature=+atomics`
# Description

Building with `target-feature=+atomics` on `wasm32-unknown-unknown` with the `v4` and `js` features currently fails due to `js_sys` being referenced but not included as a dependency. This was first noticed [here](https://github.com/bevyengine/bevy/actions/runs/13149675935/job/36695203159?pr=17689).

This is present on `main` and in 1.13.

# Replication

_On Windows_
```powershell
$env:RUSTFLAGS="-C target-feature=+atomics"; cargo +nightly check --target wasm32-unknown-unknown -Z build-std=std,panic_abort --features v4,js
```

_On Ubuntu_
```bash
RUSTFLAGS="-C target-feature=+atomics" bash -c 'cargo +nightly check --target wasm32-unknown-unknown -Z build-std=std,panic_abort --features v4,js'
```

bushrat011899:
# Objective

- Fixes #796

# Solution

- Include `js_sys` explicitly when enabling the `js` feature on `wasm32` when compiling with the `atomics` `target_feature` enabled.
- Fix possibly outdated code from vendoring

# Notes

- `js_sys` was already in the dependency graph due to `wasm-bindgen`, it just needed to be exposed when compiling under the conditions described in the attached issue.
- This is not a breaking change and could be published as 1.13.1 to fix 1.13.0 failing to compile in the aforementioned conditions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,6 +133,11 @@ jobs:
       - name: Fast RNG
         run: wasm-pack test --node -- --features "js v4 fast-rng"
 
+      - name: +atomics
+        env:
+          RUSTFLAGS: '-C target-feature=+atomics'
+        run: cargo check --target wasm32-unknown-unknown --features "$VERSION_FEATURES $DEP_FEATURES js"
+
       - name: rng-getrandom
         env:
           RUSTFLAGS: '--cfg getrandom_backend="wasm_js"'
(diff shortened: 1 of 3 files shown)


KodrAus:
Thanks @bushrat011899 :+1:

I'll get this merged and released, and rejig our CI around wasm to better catch this sort of thing in the future.

bushrat011899:
I've just copied in the CI change from #798 with you credited so this should sort out everything hopefully!

PR #808, opened 2025-02-24
Add a manual `Debug` implementation for NonNilUUid

rick-de-water:
`NonNilUuid` currently has an automatic `Debug` trait implementation, resulting in an output that is not very useful:

```
NonNilUuid(1512366075204170929049582354406559215)
```

This PR adds a manual implementation of the `Debug` trait that properly formats the uuid, matching that of the `Uuid` type.

### DIFF
--- a/src/non_nil.rs
+++ b/src/non_nil.rs
@@ -38,13 +38,13 @@ pub struct NonNilUuid(NonZeroU128);
 
 impl fmt::Debug for NonNilUuid {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", Uuid::from(*self))
+        fmt::Debug::fmt(&Uuid::from(*self), f)
     }
 }
 
 impl fmt::Display for NonNilUuid {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", Uuid::from(*self))
+        fmt::Display::fmt(&Uuid::from(*self), f)
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/non_nil.rs
```diff
@@ -33,9 +33,15 @@ use crate::{
 /// may change. It is currently only guaranteed that `NonNilUuid` and `Option<NonNilUuid>`
 /// are the same size as `Uuid`.
 #[repr(transparent)]
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
 pub struct NonNilUuid(NonZeroU128);
 
+impl fmt::Debug for NonNilUuid {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", Uuid::from(*self))
```
KodrAus, line 41: I think it would be better to write this as:

```
fmt::Display::fmt(&Uuid::from(*self), f)
```

This uses less of the formatting machinery and preserves any flags on `f`.

KodrAus (commented):
Thanks @rick-de-water! This is a good idea. I just had a comment on the impl.

src/non_nil.rs
rick-de-water: I've changed it to directly calling `Debug::fmt`, and also updated the `Display` implementation to do the same. My change was incorrect as well (should have been using `"{:?}"` instead of `"{}"`), but that should be fixed with the direct calls as well.

KodrAus (approved):
Thanks @rick-de-water! This looks good to me

PR #809, opened 2025-02-25
Support higher precision, shiftable timestamps in V7 UUIDs

### LINKED ISSUE #767: Special UUIDv7 function for microservices
Hi @KodrAus

Although fault tolerance requires that each microservice writes to its own database tables, in practice this requirement is often violated.

The implementation of UUIDv7 for PostgreSQL had to switch from Method 1 to [Method 3](https://datatracker.ietf.org/doc/html/rfc9562#name-monotonicity-and-counters) (Increased Clock Precision with 12 bits sub-millisecond timestamp fraction) to synchronize the UUIDv7s generated by different microservices for the same database table. This turned out to be simpler than the [autoincrement-like](https://postgrespro.com/docs/postgresql/16/datatype-numeric#DATATYPE-SERIAL) analogue. See the C implementation [v27-0001-Implement-UUID-v7.patch](https://www.postgresql.org/message-id/attachment/163415/v27-0001-Implement-UUID-v7.patch) of Method 3 at [the page](https://commitfest.postgresql.org/47/4388/) as a reference. The entire timestamp acts as a counter in rare case when more than about 4 identifiers per microsecond are generated.

This implementation also added the ability to offset the timestamp by a specified [interval](https://postgrespro.com/docs/postgresql/16/datatype-datetime#DATATYPE-INTERVAL-INPUT) to hide the record creation time for information security. If offset would cause the timestamp to be outside the allowed range, it should not be applied.

It would be nice to add such a special UUIDv7 function for microservices.

KodrAus:
Closes #767

This PR refactors our `V7Context` type to support timestamps of varying precision. It can be used to improve the sortability of UUIDs in distributed systems by trading some entropy for wider timestamps.

I've also added support for shifting timestamps by some specified amount in applications that want to obfuscate the time a UUID was actually generated.

Still to do:

- [x] Mask increased precision into the counter so V7 UUIDs actually use it.
- [x] Support shifting by larger amounts, or at least ensure we don't make it impossible to support in the future.
- [x] Benchmark changes and ensure we don't regress in the case where neither increased precision nor shifting are applied.

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -668,7 +668,7 @@ pub mod context {
     mod v7_support {
         use super::*;
 
-        use core::{cell::Cell, panic::RefUnwindSafe};
+        use core::{cell::Cell, panic::RefUnwindSafe, convert::TryInto};
 
         #[cfg(feature = "std")]
         static CONTEXT_V7: SharedContextV7 =
@@ -707,6 +707,8 @@ pub mod context {
         #[derive(Debug)]
         pub struct ContextV7 {
             last_reseed: Cell<LastReseed>,
+            shift: Shift,
+            sub_milli_precision_bits: u8,
             counter: Cell<u64>,
         }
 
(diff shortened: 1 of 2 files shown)


src/timestamp.rs
```diff
@@ -756,52 +841,58 @@ pub mod context {
...
-                        let counter = crate::rng::u64() & RESEED_MASK;
+                        counter = crate::rng::u64() & RESEED_MASK;
                         self.counter.set(counter);
-
-                        (counter, last_reseed.ts_seconds, last_reseed.ts_subsec_nanos)
                     }
+                };
+
+                // If we're using additional precision then mask it into the front of the counter
+                if self.additional_precision_bits != 0 {
```
KodrAus, line 885: We should be able to handle overflow more naturally here by masking into the counter when we fetch it, rather than at the end

sergeyprokhorenko:
Hi @KodrAus

Today I found [this code](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/adt/uuid.c;h=2e32592f5722407d6524e24811a5b5b88d419c96;hb=78c5e141e9c13) for the next version 18 of PostgreSQL

sergeyprokhorenko:
Hi @KodrAus

Today I found [this code](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/adt/uuid.c;h=2e32592f5722407d6524e24811a5b5b88d419c96;hb=78c5e141e9c13) for the next version 18 of PostgreSQL

PR #815, opened 2025-03-08
Mark `Uuid::new_v8` const

### LINKED ISSUE #813: Make `new_v8` const
Internally `Uuid::new_v8` uses `Builder::from_custom_bytes(buf).into_uuid()` which is `const`.

tguichaoua:
resolves #813


KodrAus (approved):
Thanks @tguichaoua! This looks good to me

PR #819, opened 2025-03-18
Added convenience implementation TryFrom<String> for std

Nahuel-M:
It's directly ported to `TryFrom<&str>`, but makes writing some macros a lot easier.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -19,6 +19,9 @@ use crate::{
     Uuid,
 };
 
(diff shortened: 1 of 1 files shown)


Nahuel-M:
@KodrAus 
The docs say to ping a maintainer. I hope this implementation makes sense from your perspective as well.

KodrAus (approved):
Thanks @Nahuel-M. This looks reasonable to me :+1:

PR #820, opened 2025-03-21
Support bincode 2

MavethGH:
This PR adds a new feature (`bincode2`) that when enabled, derives [`Encode`](https://docs.rs/bincode/latest/bincode/enc/trait.Encode.html) and [`Decode`](https://docs.rs/bincode/latest/bincode/de/trait.Decode.html) for `Uuid`. See [this section of the migration guide](https://docs.rs/bincode/latest/bincode/migration_guide/index.html#bincode-derive-and-libraries) for bincode 2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,7 @@ md5 = ["dep:md-5"]
 atomic = ["dep:atomic"]
 
 borsh = ["dep:borsh", "dep:borsh-derive"]
+bincode2 = ["dep:bincode"]
 
 # Public: Used in trait impls on `Uuid`
 [dependencies.bytemuck]
@@ -103,6 +104,14 @@ version = "2"
 [dependencies.arbitrary]
 optional = true
 version = "1.1.3"
+
+# Public: Used in trait impls on `Uuid`
+[dependencies.bincode]
+default-features = false
+features = ["derive"]
+optional = true
+version = "2.0"
+
 
 # Public (unstable): Used in `zerocopy` derive
 # Unstable: also need RUSTFLAGS="--cfg uuid_unstable" to work
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -455,6 +455,7 @@ pub enum Variant {
     feature = "bytemuck",
     derive(bytemuck::Zeroable, bytemuck::Pod, bytemuck::TransparentWrapper)
 )]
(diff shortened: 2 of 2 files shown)


KodrAus:
Hi @MavethGH 👋 I’m not familiar with `bincode` 2 yet. If `bincode` 2.x is expected to produce compatible bytes as 1.x then we should make sure we serialize to the same binary when using `serde` + `bincode` 1.x, and when using `bincode` 2.x.

MavethGH:
> If `bincode` 2.x is expected to produce compatible bytes as 1.x then we should make sure we serialize to the same binary when using `serde` + `bincode` 1.x, and when using `bincode` 2.x.

The bytes will not be the same, since `serde` includes the length of the array, while `bincode` 2.x doesn't since it's a fixed-size array. However, `bincode` 2.x can still deserialize a `Uuid` serialized that way by using [`Compat`](https://docs.rs/bincode/2.0.1/bincode/serde/struct.Compat.html) and [`config::legacy()`](https://docs.rs/bincode/latest/bincode/config/fn.legacy.html).

KodrAus:
It looks like `bincode` 2.x does include the length of the array if we serialize it as a `&[u8]` instead of as `[u8; 16]`, which is how we do it with `serde` and `bincode` 1.x just because that's the API `serde` gives us.

cc @ZoeyR @VictorKoenders who look like they might be maintaining `bincode` now, is there a general expectation that if you already support `bincode` 1.x that you'll maintain an equivalent encoding in `bincode` 2.x? Or should users expect there to be differences?

ZoeyR:
If at all possible the encodings should be equivalent.

KodrAus:
Thanks @ZoeyR. In that case, I think we should implement `Encode` and `Decode` manually, and add a test to ensure equivalence with `serde` + `bincode` 1.x. We don't necessarily need to depend on `serde` and `bincode` 1.x to do it, we can just use a pre-serialized buffer.

francis-spliced:
I worked a a simple manual implementation of Encode & Decode
https://github.com/uuid-rs/uuid/pull/821

francisbr:
I worked a a simple manual implementation of Encode & Decode
https://github.com/uuid-rs/uuid/pull/821

MavethGH:
@francisbr Thanks!

MavethGH:
Going to close this now in that case.

PR #821, opened 2025-03-27
Add support for bincode v2.0

francisbr:
Following the latest major release of Bincode, the crates now relies on it's own Trait (Encode/Decode) it switch away from serde's De/Serialize. This PR is to add a feature flag here to optionally implement these traits.

The 2.0 bincode release: https://github.com/bincode-org/bincode/releases/tag/v2.0.0


src/external/bincode_support.rs
```diff
@@ -0,0 +1,221 @@
...
+    de::{BorrowDecoder, Decoder},
+    enc::Encoder,
+    error::{DecodeError, EncodeError},
+    BorrowDecode, Decode, Encode,
+};
+use std::string::ToString;
+
+impl Encode for Uuid {
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        Encode::encode(self.as_bytes(), encoder)
```
KodrAus, line 28: Unfortunately, in order to maintain the original `serde` 1.x + `bincode` 1.x encoding, we'll need to include the redundant length bytes here too. Here's an example original format UUID we can use to test equivalence:

```
[16, 0, 0, 0, 0, 0, 0, 0, 232, 18, 203, 187, 159, 207, 72, 46, 151, 197, 95, 187, 59, 106, 126, 5]
```

KodrAus (commented):
Thanks for working on this @francisbr! As much as I'd love to fix the redundant length bytes, we need to maintain compatibility with the old `serde` and `bincode` 1.x format. If the new `bincode` derive macros have a similar concept to `serde`'s `#[with]` for customizing serialization then we could add support for the more compact format that way.

Cargo.toml
```diff
@@ -29,17 +29,17 @@ homepage = "https://github.com/uuid-rs/uuid"
 name = "uuid"
 readme = "README.md"
 repository = "https://github.com/uuid-rs/uuid"
-version = "1.16.0" # remember to update html_root_url in lib.rs
+version = "1.16.1" # remember to update html_root_url in lib.rs
```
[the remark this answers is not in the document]
francisbr: Got it, reverted!

src/external/bincode_support.rs
francisbr: I am not 100% sure I understand how that was working. The example you gave me is 24 bytes (more than the 16 or 17 with the length byte).

I added a check in the encode/decode function to check if the config passed is legacy. If so I simply prepend the '16' byte before the UUID. Is that what is needed?
You can see the `test_legacy()` test I added, when passing `config::legacy()` to the functions, 17 bytes are encoded/decoded.

Sorry if that is not what is needed and I missunderstood.

KodrAus: Yes I thought our original encoding gave us a 1 byte length, but running `bincode::serialize` on 1.x gave me what appeared to be a `usize` length prefix

src/external/bincode_support.rs
```diff
@@ -0,0 +1,288 @@
...
+    BorrowDecode, Decode, Encode,
+};
+use std::string::ToString;
+
+impl Encode for Uuid {
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        let config = encoder.config();
+
+        // Legacy serde/bincode 1.0
+        if config.endianness() == Endianness::Little
```
KodrAus, line 34: Is this configuration a new feature of `bincode` 2.x? It seems odd for us to not encode the length byte if any of these switches are set, since none of them seem related to it

src/external/bincode_support.rs
```diff
@@ -0,0 +1,288 @@
...
+        let v2_legacy_bytes = bincode::encode_to_vec(&V2Container(uuid), config::legacy()).expect(
+            &format!("Should have been able to encode V2Container({uuid_str}) & legacy config."),
+        );
+        let legacy_bytes = bincode::encode_to_vec(&LegacyContainer(uuid), config::standard())
+            .expect(&format!(
+                "Should have been able to encode LegacyContainer({uuid_str})."
+            ));
+
+        assert_eq!(legacy_bytes, v2_legacy_bytes);
+        assert_eq!(17, v2_legacy_bytes.len());
```
KodrAus, line 158: Looks like you're getting 17 bytes out of `bincode` instead of 24. I wonder if the `config::standard()` is using variable-length integers instead of fixed size. I think just to be sure we could include `bincode` 1.x as a dev dependency as well with:

```toml
[dev-dependencies.bincodev1]
package = "bincode"
version = "1"
```

And use `bincode::serialize` as our legacy format.

Sorry for being so pedantic on this, I just don't want to accidentally ship an unintentional format change that we're stuck with, and that creates confusion for anyone upgrading `bincode`.

francisbr: Nice suggestion, I didn't think of that! Don't be sorry, I totally understand the goal! The test make much more sense like that I think.

src/external/bincode_support.rs
francisbr: Yes configuration is a new feature of v2. `config::standard()` uses variable length encoding vs legacy uses fixed length (usized as you mentionned). You can see in the test function the v2 bincode function requires a config.
And you can read Bincode's migration guide here: https://docs.rs/bincode/2.0.1/bincode/migration_guide/index.html

I refactored to use only this config flag, always encoding the length but as a `usize` in fixed length encode case & as a `u8` in the case of variable length encoding.

francisbr: And to give more explanation, I was conditionally encoding the leading length byte, because since bincode requires this config to be compatible, we could take the opportunity to remove the unecessary leading byte if someone migrates to v2 + standard config.

But, obviously, I am not opposed to keeping it for sure, just wanted to save that byte of space! 😅

KodrAus (approved):
Thanks @francisbr! This looks good to me.

KodrAus:
Looks like we've got a few build failures here:

```
failures:

---- external::bincode_support::bincode_tests::test_decode_failure stdout ----

thread 'external::bincode_support::bincode_tests::test_decode_failure' panicked at src/external/bincode_support.rs:221:18:
Unexpected error
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- external::bincode_support::bincode_tests::test_decode_non_nil_uuid stdout ----

thread 'external::bincode_support::bincode_tests::test_decode_non_nil_uuid' panicked at src/external/bincode_support.rs:239:9:
assertion `left == right` failed
  left: 16
 right: 17
```

src/external/bincode_support.rs
```diff
@@ -0,0 +1,259 @@
...
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        self.as_uuid().encode(encoder)
+    }
+}
+
+impl Encode for Braced {
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        self.as_uuid().encode(encoder)
+    }
+}
```
loarca, line 69: Wouldn't these differ from the `serde` impl?

`serde` would output strings for these, therefore I believe bincode v1 encodes a string rather than the bytes.

Do we care, though?

src/external/bincode_support.rs
```diff
@@ -0,0 +1,259 @@
...
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        if encoder.config().int_encoding() == IntEncoding::Fixed {
+            usize::encode(&16, encoder)?;
+        } else {
+            u8::encode(&16, encoder)?;
+        }
+
+        <&[u8; 16]>::encode(&self.as_bytes(), encoder)?;
+
+        Ok(())
```
loarca, line 37: I believe this is sufficient and delegates the int encoding implementation to bincode itself
```suggestion
        <&[u8]>::encode(&self.as_bytes().as_slice(), encoder)
```

src/external/bincode_support.rs
```diff
@@ -0,0 +1,259 @@
...
+    fn borrow_decode<D: BorrowDecoder<'de, Context = Context>>(
+        decoder: &mut D,
+    ) -> Result<Self, DecodeError> {
+        let uuid = Uuid::borrow_decode(decoder)?;
+        NonNilUuid::try_from(uuid).map_err(|e| DecodeError::OtherString(e.to_string()))
+    }
+}
+
+#[cfg(test)]
+mod bincode_tests {
```
loarca, line 112: Besides testing the roundtrip of encode->decode, we should probably also test the output binary representation of encode. Just to ensure we know what to expect from a `Uuid` and its variants (`Simple`, etc)

src/external/bincode_support.rs
KodrAus: Ah well spotted! We definitely should maintain compatibility on these adapter types too

src/external/bincode_support.rs
KodrAus: Leaning on the impl provided by `bincode` itself seems like a good idea. That way we should naturally respect any new options it introduces that might affect encoding that end users may want.

Issue #775, opened 2024-12-12
Hash faster by not using derive


jrose-signal:
Uuid currently uses `derive(Hash)`, but this calls through to an array's default implementation, which in turn calls to the implementation for *slices*, which includes a length prefix and goes element by element. Instead, it could just call `state.write(&self.0)` directly.

kkpo4553:
Hello, we tried to solve the issue.

This is what we did:

Remove the derive(Hash) attribute from Uuid and implement a custom Hash trait to optimize hashing by directly writing the bytes to the hash state.

You can review changes in this commit: https://github.com/kkpo4553/uuid-rs-uuid-775/commit/588a8130bc930be2b3732eea23086c5044a3982e.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

KodrAus:
The `Hash` trait doesn’t guarantee stability of hashes unless you have control over both the hasher and type being hashed, so I think we should be able to make this change.

It’s technically unrelated, but what we can’t unfortunately do is change how we compute hashes in v1, v3, or v5 UUIDs, or else we’d start computing a different UUID for the same input.

jrose-signal:
That's fine, those already pass the bytes directly. (And they are fully specified by the UUID standard, so they're either correct or wrong as they are.)

PR #824, opened 2025-04-03
Replace derive(Hash) with manual impl in Uuid

diopoex:
My attempt to solve #775.

I am not sure about the following two things:
- If I imported core::hash::{Hash, Hasher} in an appropriate style.
- If the other derive(Hash) should also be removed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -231,6 +231,7 @@ mod parser;
 pub mod fmt;
 pub mod timestamp;
 
+use core::hash::{Hash, Hasher};
 pub use timestamp::{context::NoContext, ClockSequence, Timestamp};
 
 #[cfg(any(feature = "v1", feature = "v6"))]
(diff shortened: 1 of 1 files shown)


diopoex:
@KodrAus, its done i believe, but iam not shure if the other uses of derive(Hash) should also be removed.

kinggoesgaming:
Hi @diopoex 🖐🏼

I think for starters, this should be good.

If this implementation change does not cause issues (**it shouldn't**, but you never know), another PR can do it for the rest.

diopoex:
Hey @kinggoesgaming 

Iam open to do it for where it makes sense so that the issue is completed and that we do not have an inconsistent solution.

But lets see!

KodrAus (approved):
Thanks @diopoex! This looks good to me. The build failure here is unrelated; I fixed it up in #825

PR #828, opened 2025-05-23
Add `wasm32v1-none` Support

bushrat011899:
# Objective

`uuid` has support for `no_std`, and support for `wasm32-unknown-unknown`, but it currently _doesn't_ have support for the alternative `wasm32v1-none`. Attempting to compile `uuid` with the `v4` and `js` features on `wasm32v1-none` causes a compilation error due to `getrandom`.

## Solution

Based on [`wasm32v1-none`](https://doc.rust-lang.org/beta/rustc/platform-support/wasm32v1-none.html#conditionally-compiling-code) and [`wasm32-unknown-unknown`](https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-unknown-unknown.html#conditionally-compiling-code) documentation, and [this compilation failure condition](https://github.com/rust-random/getrandom/blob/2cf460252aaa897453801965e3e7c1cbf301d04d/src/backends.rs#L170) in `getrandom`, the configuration condition `uuid` should use for possibly-web Wasm is:

> all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))

That is, 32-bit Wasm with either an unknown OS (`wasm32-unknown-unknown`) or none (`wasm32v1-none`). The `target_vendor` condition is omitted based on the recommendations from the official Rust documentation linked above.

---

## Notes

- I have also removed the implicit `std` features from `js-sys` and `wasm-bindgen`, and moved them into `uuid/std`. This may not be required, as `wasm32-unknown-unknown` still compiles without these `std` features enabled, but doing so reduces the chance of downstream issues.
- I've added a CI task "MVP Web" to ensure this particular target continues to compile going forward. Note that due to development dependency feature unification on older versions of Rust, you may need to use `-Z avoid-dev-deps` to test the target locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
         uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
 
       - name: Install Rust toolchain
-        run: rustup update nightly && rustup target add --toolchain stable wasm32v1-none
+        run: rustup update nightly && rustup target add --toolchain nightly wasm32v1-none
 
       - name: No features
         run: cargo +nightly build -Z avoid-dev-deps --target wasm32v1-none --no-default-features
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,3 +250,19 @@ jobs:
 
       - name: Powerset
         run: cargo hack check --each-feature -Z avoid-dev-deps
+
+  wasm32v1-none:
+    name: Build / MVP Web
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
+
+      - name: Install Rust toolchain
+        run: rustup update nightly && rustup target add --toolchain stable wasm32v1-none
+
+      - name: No features
+        run: cargo +nightly build -Z avoid-dev-deps --target wasm32v1-none --no-default-features
+
+      - name: Version features
+        run: cargo +nightly build -Z avoid-dev-deps --target wasm32v1-none --no-default-features --features "js v1 v3 v4 v5 v6 v7 v8 serde"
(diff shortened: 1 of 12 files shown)


KodrAus:
Thanks for working on this @bushrat011899! Just so I follow properly, we treat the `wasm32v1-none` target the same as `wasm32-unknown-unknown` with respect to our source of randomness? That is, if you enable the `js` feature of `uuid` you'll get our internal RNG support, but if you enable `rng-rand` or `rng-getrandom` you'll get theirs?

bushrat011899:
> Just so I follow properly, we treat the `wasm32v1-none` target the same as `wasm32-unknown-unknown` with respect to our source of randomness? That is, if you enable the `js` feature of `uuid` you'll get our internal RNG support, but if you enable `rng-rand` or `rng-getrandom` you'll get theirs?

Correct, the only difference between `wasm32v1-none` and `wasm32-unknown-unknown` is `wasm32v1-none` does not have a pre-built `std` implementation, and it explicitly uses the `mvp` `target-cpu`, which provides much better backwards compatibility with older web browsers. There's some further details in the [documentation](https://doc.rust-lang.org/beta/rustc/platform-support/wasm32v1-none.html) for the target. But the general plan is for `wasm32-unknown-unknown` to be the "bleeding edge" of Wasm, implementing all features and functionality in the spec that is "generally available", while the `wasm32vX-none` target(s) will strictly adhere to the `X`'th [Web Assembly Core Spec](https://www.w3.org/TR/wasm-core-1/) for a more stable level of compatibility.

The `no_std` component here is a slightly more controversial choice, but is motivated by the fact that `wasm32-unknown-unknown` simply panics when certain elements from the `std` library are used, since they just have stub implementations. Making `wasm32v1-none` `no_std` avoids those footguns entirely, and is a better direction for Wasm on the web (IMO, at least until `std` can use the JS runtime and assume its existence somehow)

KodrAus (approved):
Thanks @bushrat011899! This looks great to me.

Issue #534, opened 2021-10-19
Implement `Serialize` & `Deserialize` for `Hyphenated`


thomaseizinger:
**Is your feature request related to a problem? Please describe.**

Would like to derive `Serialize` on a newtype that wraps `Hyphenated`.

**Describe the solution you'd like**

`derive(Serialize, Deserialize)` to just work.

**Is it blocking?**
-

**Describe alternatives you've considered**

Manually implement `Serialize` on my newtype. Works but is verbose.

**Additional context**
-

**Other**
-

thomaseizinger:
> Hey @thomaseizinger 👋
> 
> I think this is just a missing bit of functionality 🙂 Would you be happy to PR a `Serialize` implementation for our format adapters?

Sorry, I've been on vacation and you beat me to it, otherwise would have happily sent a PR!

Thank you!

Kek5chen:
Hello, excuse me but I don't think this is closed. #546 implemented Serialize, but not Deserialize on the Format Variants.

PR #832, opened 2025-07-06
Implement deserialize for format uuid types

Kek5chen:
Finished the work started in #534.

Implemented deserialize for format types
Refactors serialize statements into a macro call for brevity.

### DIFF
--- a/src/external/serde_support.rs
+++ b/src/external/serde_support.rs
@@ -41,29 +41,20 @@ impl Serialize for NonNilUuid {
     }
 }
 
-impl Serialize for Hyphenated {
-    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        serializer.serialize_str(self.encode_lower(&mut Uuid::encode_buffer()))
-    }
+macro_rules! serialize_uuid_proxy {
+    ($type:ty) => {
+        impl Serialize for $type {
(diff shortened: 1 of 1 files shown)


src/external/serde_support.rs
```diff
@@ -138,6 +129,24 @@ impl<'de> Deserialize<'de> for Uuid {
...
 }
 
+macro_rules! deserialize_uuid_proxy {
+    ($type:ty) => {
+        impl<'de> Deserialize<'de> for $type {
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                Ok(Uuid::deserialize(deserializer)?.into())
```
KodrAus, line 139: If you're trying to deserialize a UUID into one of these proxy types, I think we should only accept a UUID in its format. We should be able to make the visitor used in `Uuid`'s deserialize impl generic and add appropriate `FromStr` implementations to the proxy types. Then the bound on that visitor would be any `T: FromStr + From<Uuid>`.

I'd be happy to sketch it out a bit further if that would be helpful 🙂

KodrAus (commented):
Thanks for the ping @Kek5chen!

src/external/serde_support.rs
Kek5chen: Alright, I can change it to have that behavior

PR #833, opened 2025-07-08
Remove some unsafe; stabilize zerocopy

### LINKED ISSUE #588: Tracking issue for zerocopy
This issue is a place to keep anybody using the unstable `zerocopy` feature up-to-date on any changes being made to it.

In order to use `zerocopy`, you need to opt-in not just to the feature, but to the `uuid_unstable` RUSTFLAGS configuration too. We can make breaking changes to `zerocopy` at any point, but still want to be considerate of anybody making use of it, so will try keep open about any version bumps or other breakage that's coming.

joshlf:
Remove a number of `unsafe` blocks, replacing them with uses of zerocopy. In order to do this, we stabilize zerocopy as a (non-optional) dependency.

Closes #588

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,13 +104,8 @@ version = "2"
 optional = true
 version = "1.1.3"
 
-# Public (unstable): Used in `zerocopy` derive
-# Unstable: also need RUSTFLAGS="--cfg uuid_unstable" to work
-# This feature may break between releases, or be removed entirely before
-# stabilization.
-# See: https://github.com/uuid-rs/uuid/issues/588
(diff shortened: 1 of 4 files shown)


KodrAus:
Thanks for the PR @joshlf! I'm not immediately in favour of adding a non-optional dependency on `zerocopy` to remove a handful of unsafe blocks. Since it's also still technically unstable we're not able to accept a public dependency on it here anyways.

As an aside, it would be really nice to see some support for the common pattern of transmuting values of `#[repr(transparent)] NewType(T)` to `T` without needing an `unsafe` block in the language or standard library.

PR #846, opened 2025-11-17
feat: add miniserde support

ymelois:
adds support for [miniserde](https://crates.io/crates/miniserde).


KodrAus:
Thanks for the PR @sehnryr! We unfortunately can't support `miniserde` directly in `uuid` because it's unstable. I think your best bet here would be to create a newtype wrapper over `Uuid` that implements `miniserde` traits pretty much exactly as you have them here.

PR #852, opened 2025-12-21
feat: impl core::error::Error for Error

XAMPPRocky:
Thís will bump the requirement up to 1.81 but this brings a huge improvement to no-std environments.


KodrAus:
> I don't have any skin in this, but could the new MSRV-aware resolver help to mitigate the MSRV incompatibility for users using older Rust versions?

Quite possibly. I'd be interested to see how the rules change with more tools in this area. I have my personal perspective on this, which is that MSRV's are a misfeature and shouldn't exist, but for libraries I try maintain policies that minimize the impact on the widest number of users. I do agree with @XAMPPRocky that `std::core::error` is a big enough feature that it's worth considering a bump for, like const generics were a few years ago.

PR #858, opened 2026-02-06
Update getrandom to 0.4

KodrAus:
The [changelog for `getrandom`](https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md#040---2026-02-02) suggests the major change here is a bump of the MSRV. I've already been on the fence about doing this for #852, so if it otherwise remains compatible, I think we can accept the bump and move all of these things forwards.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ name = "uuid"
 readme = "README.md"
 repository = "https://github.com/uuid-rs/uuid"
 version = "1.20.0" # remember to update html_root_url in lib.rs
-rust-version = "1.81.0"
+rust-version = "1.85.0"
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


KodrAus:
Also to check: whether `Uuid::new_v4()` on Windows is better without `rand` now that `getrandom` is using a different Windows API. If so, I'll deprecate that feature flag too.

KodrAus:
Looks like `rand` itself doesn't depend on `getrandom` `0.4` yet, so I suspect this is blocked until its `0.10.0` release comes out.

sabify:
Thanks!

Can you please cut a new release?

Cargo.toml
hanyuone: Was this missed in the PR, should this be `0.10`?

PR #863, opened 2026-03-04
Default to rand 0.10

haxtibal:
rand 0.10.0 was [released 2026-02-08](https://github.com/rust-random/rand/releases/tag/0.10.0) and can be used without code change. Default to it while still allowing 0.9.x for wider compatibility.


haxtibal:
> Thanks for the PR @haxtibal! This was simply missed during the update to `getrandom` `0.4`.

Thx, I see. Changed commit message accordingly and applied your suggestion.

PR #878, opened 2026-04-15
fix: Timestamp::from_gregorian deprecation note

aznashwan:
Fix deprecation note for `Timestamp::{from_gregorian,from_rfc4122}` from erroneously recommending to use `Timestamp::from_gregorian` instead of the correct `from_gregorian_time`.

This leads to an especially insidious-looking deprecation warning which left me in disbelief:

Screenshot 2026-04-15 at 09 32 27


KodrAus (approved):
Ah thanks @aznashwan!

PR #886, opened 2026-06-09
Fix up parser panic on empty input

### LINKED ISSUE #885: Panic parsing invalid UUID input "{}"
version = "1.23.2"

Using `Uuid::parse_str("{}");` causes 
```
thread 'main' (75) panicked at /playground/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/uuid-1.23.2/src/error.rs:75:21:
attempt to subtract with overflow
```
Reproducible:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b73d1ba4712236f57ac98ffb9d1c940e

KodrAus:
Closes #885 

I've updated our fuzzing infra along the way and will make it run for a short time in CI from now on. The parsers are quite simple, so it only takes a few seconds for AFL to map its state space and find crashes. This should help us catch these kinds of regressions faster in the future.


KodrAus:
I haven't actually fixed the bug yet, I just want to watch it crash and burn in CI first.

KodrAus:
Gotcha:

```
repro: "/home/runner/work/uuid/uuid/fuzz/target/parse/default/crashes/id:000001,sig:06,src:000003,time:1955,execs:63478,op:havoc,rep:8"
repro: [123, 125]
repro: "{}"
thread 'tests::repro' (7216) panicked at /home/runner/work/uuid/uuid/src/error.rs:75:21:
attempt to subtract with overflow
```

PR #889, opened 2026-06-24
document fixes

frostyplanet:
1. Some function under feature flag (for example <https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.now_v1> ) does not have labels about feature condition
2. Fix broken link during doc compilation

### DIFF
--- a/src/external/serde_support.rs
+++ b/src/external/serde_support.rs
@@ -360,7 +360,7 @@ pub mod compact {
     }
 }
 
-/// Serialize a [`Uuid`] as [`uuid::fmt::Simple`].
+/// Serialize a [`Uuid`] as [`uuid::fmt::Simple`]([crate::fmt::Simple]).
 ///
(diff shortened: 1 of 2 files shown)


KodrAus (approved):
Thanks @frostyplanet!

src/external/serde_support.rs
```diff
@@ -360,7 +360,7 @@ pub mod compact {
     }
 }
 
-/// Serialize a [`Uuid`] as [`uuid::fmt::Simple`].
+/// Serialize a [`Uuid`] as [`uuid::fmt::Simple`]([crate::fmt::Simple]).
```
Kriskras99, line 363:
```suggestion
/// Serialize a [`Uuid`] as [`uuid::fmt::Simple`](crate::fmt::Simple).
```
The brackets should not be included.

frostyplanet: sorry, did not notice, but the PR is merged,  open another one

PR #892, opened 2026-07-08
feat(fmt): support encoding into MaybeUninit buffers

### LINKED ISSUE #884: Add `[MaybeUninit<u8>]` variants of `encode_*` functions
As discussed [here](https://www.reddit.com/r/rust/comments/1tqgz20/comment/oojtuyf/), it'd be nice to provide encode functions that accept `MaybeUninit` buffers instead, for performance-critical applications that want to avoid unnecessary allocations and copies.

Example where this would be useful:
```rust
use std::sync::Arc;
use uuid::Uuid;

fn new_uuid() -> Arc<str> {
	let mut id =
		Arc::<[u8]>::new_uninit_slice(uuid::fmt::Hyphenated::LENGTH);
	// FIXME: Use `Arc::get_mut_unchecked` when it becomes stable.
	// SAFETY: Only one reference to `id` exists.
	let buf = unsafe { Arc::get_mut(&mut id).unwrap_unchecked() };

	// `encode_lower_uninit` must guarantee it won't ever read from `buf`.
	Uuid::new_v4().as_hyphenated().encode_lower_uninit(buf);

	// SAFETY: `encode_lower_unit` filled `buf` with valid ASCII bytes.
	unsafe { Arc::from_raw(Arc::into_raw(id) as *const str) }
}
```

As an alternative to such variants, the docs of encode_* functions could guarantee they won't ever read from the buffers, but this would force the caller into making some very uncomfortable casts.

weifanglab:
## Summary

- Add `encode_lower_uninit` and `encode_upper_uninit` methods for `Hyphenated`, `Simple`, `Urn`, and `Braced`.
- Allow UUID formatters to encode directly into `&mut [MaybeUninit<u8>]`.
- Reuse the new uninit encoding path from the existing `encode_lower` / `encode_upper` methods.
- Closes #884.

## Details

The new methods initialize the returned prefix of the provided buffer and leave bytes outside of the returned subslice untouched. This follows the API shape discussed in #884 while keeping the existing initialized-buffer APIs available and unchanged.

## Testing

- `rustfmt --check src/fmt.rs`
- `cargo test --lib`
- `cargo test --all-features --lib`
- `cargo test --all-features --doc`
- `cargo test --all-features --tests`

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -11,7 +11,7 @@
 
 //! Adapters for alternative string formats.
 
-use core::str::FromStr;
+use core::{mem::MaybeUninit, slice, str::FromStr};
 
 use crate::{
     std::{borrow::Borrow, fmt, str},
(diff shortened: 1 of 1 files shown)


src/fmt.rs
```diff
@@ -230,59 +230,124 @@ const fn format_hyphenated(src: &[u8; 16], upper: bool) -> [u8; 36] {
...
+
+    // SAFETY: The encoded buffer is fully initialized and ASCII encoded.
+    unsafe { assume_init_ascii_mut(buf) }
+}
+
+#[inline]
+fn write_bytes(dst: &mut [MaybeUninit<u8>], src: &[u8]) {
+    debug_assert_eq!(dst.len(), src.len());
+
+    for (dst, src) in dst.iter_mut().zip(src) {
```
KodrAus: This optimizes reasonably well, but since we're already dealing with `MaybeUninit` it wouldn't be out of place to use `ptr::copy_nonoverlapping`, which optimizes slightly better:

```rust
unsafe {
    std::ptr::copy_nonoverlapping(src.as_ptr(), dst.as_mut_ptr() as *mut u8, src.len());
}
```

weifanglab: @KodrAus   Good point, thanks! 

I updated `write_bytes` to use `core::ptr::copy_nonoverlapping` and kept the safety invariant local to the helper. Please review it again.

KodrAus (approved):
Thanks @weifanglab!

PR #894, opened 2026-07-10
perf: Optimize UUID hex parsing and formatting

geeknoid:
Replace the byte-at-a-time lookup-table hex encode/decode in the parser and formatter with branchless arithmetic. Both stay const-evaluable and free of unsafe, and the loops now auto-vectorize (e.g. SSE2 on x86_64, NEON on aarch64), roughly halving instruction counts on the hot paths.

Benchmarks below compare before/after using the gungraun (Callgrind instruction counts) and libtest (wall-clock) harnesses added here.
```
  Benchmark                  Instructions       Time (ns/iter)
                             before  after      before  after
  ------------------------  -------  -----      ------  -----
  parse_random                  261    112          40     14
  parse_nil                     261    113          40     14
  parse_random_hyphenated       271    134          37     29
  parse_nil_hyphenated          271    134          37     29
  parse_urn                     290    152          38     30
  encode_simple                 131     61        12.4    0.7
  encode_hyphen                 139     81         6.6    1.8
  encode_urn                    152    127         7.1    2.0
  Display (simple)              296    236        25.0   16.5
  Display (hyphenated)          304    252          26     22
  Display (urn)                 303    254          29     27
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -203,6 +203,25 @@ version = "1.0.52"
 [dev-dependencies.rustversion]
 version = "1"
 
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies.gungraun]
+version = "=0.19.2"
+
+[[bench]]
+name = "gungraun_format_str"
+harness = false
+
+[[bench]]
+name = "gungraun_parse_str"
+harness = false
+
+[[bench]]
+name = "gungraun_v4"
+harness = false
+
+[[bench]]
+name = "gungraun_v7"
+harness = false
+
 [workspace]
 members = [
     "rng",
--- a/benches/gungraun_format_str.rs
+++ b/benches/gungraun_format_str.rs
@@ -1,0 +1,83 @@
+// Gungraun benchmarks require the `gungraun-runner` binary and Valgrind, which
+// aren't available on wasm32 targets, so the harness is compiled out there.
+#[cfg(not(target_arch = "wasm32"))]
+mod bench {
+    use std::hint::black_box;
+    use std::io::Write;
+
+    use gungraun::prelude::*;
+    use uuid::Uuid;
+
+    fn setup_uuid() -> Uuid {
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -203,6 +203,25 @@ version = "1.0.52"
 [dev-dependencies.rustversion]
 version = "1"
 
+[target.'cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))'.dev-dependencies.gungraun]
```
KodrAus: I wasn't familiar with `gungraun`, it looks interesting 👀

KodrAus (approved):
Thanks for working on this @geeknoid! We previously had looked at manually vectorized routines but didn't keep them here just because I didn't want to complicate the project or our CI too much to maintain them or try port them across ISAs. If this autovectorizes nicely on x86 and ARM then that's a big win.

KodrAus:
Looks like we'll need to cfg `gungraun` out on WASM:

```
error[E0277]: the trait bound `OsString: serde::Serialize` is not satisfied
    --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gungraun-runner-0.19.2/src/api.rs:1110:44
     |
1110 |   #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
     |                                              ^^^^^^^^^ the trait `Serialize` is not implemented for `OsString`
...
1118 | /     /// The environment variables to set or pass through to the binary
1119 | |     pub envs: Vec<(OsString, Option<OsString>)>,
     | |_______________________________________________- required by a bound introduced by this call
```

Cargo.toml
geeknoid: Yes, it's a game changer. You get deterministic benchmark results instantaneously.

geeknoid:
@KodrAus The WASM build issue should be fixed. I don't know how to trigger your CI though...

PR #899, opened 2026-08-13
Fix non-ASCII character handling in parse diagnostics

### LINKED ISSUE #898: Non-ASCII characters can be misclassified in detailed parse errors
## Description

`Uuid::parse_str` may produce an incorrect detailed diagnostic for some
non-ASCII characters.

`InvalidUuid::into_err` iterates over Unicode `char`s, casts each one to
`u8`, and then checks whether the resulting byte is an ASCII hexadecimal
digit or hyphen:

```rust
let byte = character as u8;

match (format, byte.to_ascii_lowercase()) {
    (_, b'0'..=b'9' | b'a'..=b'f') => (),
    // ...
}
```

Casting a char to u8 truncates its Unicode scalar value. A non-ASCII character can therefore have a low byte that matches a valid ASCII UUID character.
For example, U+0130 (İ) has 0x30 as its low byte, which matches the ASCII character 0.

## Reproduction

```rust
let error = uuid::Uuid::parse_str("\u{0130}").unwrap_err();
println!("{error}");
```
The current result is:

`invalid length: found 2`

The expected diagnostic is:

`invalid character: found `İ` at 0`

This does not cause the parser to accept an invalid UUID. It only affects the detailed user-facing error returned by `parse_str`.

## Proposed fix

Classify the original char directly using char::is_ascii_hexdigit() instead of converting it to u8, and add a regression test for U+0130.

questfever:
Closes https://github.com/uuid-rs/uuid/issues/898

## Summary

Fix detailed UUID parse diagnostics for non-ASCII characters whose low byte matches a valid ASCII UUID character.

`InvalidUuid::into_err` previously cast each Unicode `char` to `u8` before classifying it. This lossy conversion could misclassify characters such as `U+0130` (`İ`) as ASCII hexadecimal digits.

The diagnostic logic now checks the original character with `char::is_ascii_hexdigit()`, avoiding the truncating conversion.

A regression test verifies that `U+0130` produces:

```text
invalid character: found `İ` at 0
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -104,17 +104,15 @@ impl<'a> InvalidUuid<'a> {
         let mut group_bounds = [0; 4];
 
         for (index, character) in input_str[bounds.clone()].char_indices() {
-            let byte = character as u8;
-
-            match (format, byte.to_ascii_lowercase()) {
-                (_, b'0'..=b'9' | b'a'..=b'f') => (),
-                (RequestedUuid::Simple, b'-') => {
+            match (format, character) {
+                (_, character) if character.is_ascii_hexdigit() => (),
+                (RequestedUuid::Simple, '-') => {
                     return Error(ErrorKind::ParseChar {
                         character: '-',
                         index: index + bounds.start,
                     })
                 }
-                (_, b'-') => {
+                (_, '-') => {
                     if format == RequestedUuid::Any {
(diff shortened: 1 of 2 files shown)


KodrAus (approved):
Thanks @questfever!

Issue #893, opened 2026-07-08
The `compact` custom serde representation is not particularly compact


keks:
When encoding binary data with serde, you can't just provide a `&[u8]` or `[u8; N]`, because that will just encode as a sequence of numbers. You need to use `serialize_bytes` et al., and the `serde_bytes` makes that pretty easy.

For CBOR this means that the `compact` representation produces 32 bytes instead of 17 (a tag + the bytes).

For JSON, this is especially weird, because the plain encoding is  38 bytes long (32 hex + 2 quotes + 4 dashes), but the "compact" one is 58 long (but that depends on the uuid, because it uses decimal representation.

Some examples I tested:

```
UUID: 77459ff3-c409-4d6a-a0a3-8b51ba3b56cf

serde_cbor + serde_bytes            17 bytes  5077459ff3c4094d6aa0a38b51ba3b56cf
serde_cbor + uuid::serde::compact   32 bytes  9018771845189f18f318c409184d186a18a018a3188b185118ba183b185618cf
ciborium  + serde_bytes             17 bytes  5077459ff3c4094d6aa0a38b51ba3b56cf
ciborium  + uuid::serde::compact    32 bytes  9018771845189f18f318c409184d186a18a018a3188b185118ba183b185618cf
serde_json + plain Uuid             38 bytes  "77459ff3-c409-4d6a-a0a3-8b51ba3b56cf"
serde_json + uuid::serde::compact   58 bytes  [119,69,159,243,196,9,77,106,160,163,139,81,186,59,86,207]
```

KodrAus:
Thanks for pointing this out @keks. The `compact` encoding is a bit of a legacy one we've had for a long time that was originally targeting encodings like `bincode` that encode arrays of numbers inline.

I think we could benefit from a `uuid::serde::bytes` module that used `serialize_bytes`. It would also give you some idea that there was something different about `compact` if you came to the docs expecting `compact` to be bytes.

PR #902, opened 2026-08-19
Add a serde::bytes module that encodes a Uuid as a byte string

ChrisJr404:
This picks up #893. The compact module serializes a Uuid as a [u8; 16] array, so formats that distinguish byte strings from sequences (CBOR, MessagePack) end up encoding it as 16 individually tagged integers instead of one byte string. In the CBOR example from the issue that is 32 bytes where serialize_bytes would take 17.

This adds a serde::bytes sibling module that mirrors the compact one but goes through serialize_bytes/deserialize_bytes, reusing the existing UuidBytesVisitor for the deserialize side. It is purely additive and leaves compact untouched, so nothing changes for existing users. As KodrAus noted on the issue, having a dedicated bytes module also makes it clearer that compact is the array encoding rather than a byte-string one.

### DIFF
--- a/src/external/serde_support.rs
+++ b/src/external/serde_support.rs
@@ -360,6 +360,86 @@ pub mod compact {
     }
 }
 
+/// Serialize a [`Uuid`] as a byte string.
+///
+/// Unlike [`compact`](crate::serde::compact), which serializes a [`Uuid`] as a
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks @ChrisJr404! This looks like a reasonable addition to me.

PR #904, opened 2026-08-25
Add ContextV7::with_additional_precision_bits

### LINKED ISSUE #814: Make higher precision v7 timestamps use 10 bits on platforms with microsecond precision
See: https://github.com/uuid-rs/uuid/pull/809#issuecomment-2704835750

In PostgreSQL, UUID V7 generation uses the `rand_a` field for an extra 12 bits of sub-millisecond timestamp precision. Some platforms don't generate timestamps with enough precision to fill that 12 bits though. In those cases, we can instead fill 10 bits, and leave another 2 free for random data.

This should be fairly straightforward to support in `ContextV7`, since additional precision is already pluggable and not fixed at 0 or 12 bits.

cc @sergeyprokhorenko

ChrisJr404:
Closes #814.

The `Precision` machinery already supports any bit count, so this just exposes it. `with_additional_precision_bits(bits)` lets callers on platforms with less than nanosecond resolution pick how many bits of the counter to spend on sub-millisecond precision. A microsecond clock can use 10 bits and keep the low 2 bits of `rand_a` random, as PostgreSQL v18 does.

`bits` is capped at 12 to match the size of `rand_a`, and `with_additional_precision()` now just calls this with 12 so the behavior is identical. Added a test that checks the 10-bit layout and that the two methods agree at 12 bits.

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -794,8 +794,21 @@ pub mod context {
             /// by trading a small amount of entropy for better counter synchronization. Note that the counter
             /// will still be reseeded on millisecond boundaries, even though some of its storage will be
             /// dedicated to the timestamp.
-            pub fn with_additional_precision(mut self) -> Self {
-                self.precision = Precision::new(12);
+            pub fn with_additional_precision(self) -> Self {
+                self.with_additional_precision_bits(12)
+            }
+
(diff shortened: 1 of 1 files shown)


KodrAus (approved):
Thanks @ChrisJr404! This looks good to me.

PR #907, opened 2026-09-02
Seat the v7 counter below the version nibble

lenamonj:
`Uuid::new_v7` shifts the counter so its most significant bit lands at bit 127 of the 128-bit value it hands to `Builder::from_unix_timestamp_millis`, but that builder spends the top four bits of the first payload byte on the version nibble. The top four bits of any counter are therefore discarded.

Measured at `cdc96a8`, from a scratch crate:

- with a 42-bit counter, setting any of bits 38-41 gives the same `rand_a`/`rand_b` as a counter of zero; a counter of all ones gives `rand_a = 0xff3`, not `0xfff`
- a counter stepping from `0x3fffffffff` to `0x4000000000` produces a UUID that sorts lower
- with `ContextV7::new().with_additional_precision()` the discarded bits are the sub-millisecond ones, so two UUIDs drawn from one context at two instants inside the same millisecond can sort backwards

Change: seat the counter at bit 123 (the top of `rand_a`), open the two-bit variant gap only when the counter reaches past `rand_a`, and keep the 74 most significant bits of a wider counter. Two tests: every bit of a 42-bit counter reaches the UUID and a carry sorts upwards; additional precision sorts within a millisecond.

Verified on a fresh clone at `cdc96a8`: the two tests fail before the change and pass after; `cargo test --all-features`, `--doc`, `cargo hack test --lib --each-feature`, `--all-features`, the examples crate and the 1.85.0 smoke-test build are green. wasm not run here.

Found by an automated audit loop (Claude); patch and description reviewed by me.

### DIFF
--- a/src/v7.rs
+++ b/src/v7.rs
@@ -81,10 +81,10 @@ impl Uuid {
         // so the counter is placed directly below the version nibble and shifted
         // around the variant:
         //
-        // bit 127                                                            bit 48
-        // | ver (4) | rand_a (12) | var (2) |          rand_b (62)         | ...
+        // bit 127                                                       bit 48
+        // | ver (4) |    rand_a (12)    | var (2) |       rand_b (62)        | ...
         //           |<- counter <= 12 ->|
-        //           |<----- counter > 12: shifted by 2 over the variant ---->|
+        //           |<---- counter > 12: shifted by 2 over the variant ---->|
         const RAND_A_BITS: u32 = 12;
         const PAYLOAD_BITS: u32 = RAND_A_BITS + 62;
 
--- a/src/v7.rs
+++ b/src/v7.rs
@@ -33,8 +33,9 @@ impl Uuid {
     /// # Counter treatment
     ///
     /// This method accepts a [`Timestamp`] which may include a counter value.
-    /// Only up to 74 bits of the counter value will be used when constructing
-    /// the UUID. Any unused counter bits will be filled with random data.
+    /// The 74 most significant bits of the counter value are retained when
+    /// constructing the UUID, and the rest is filled with random data. Avoid
+    /// using a counter wider than 74 bits.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


src/v7.rs
```diff
@@ -74,51 +78,44 @@ impl Uuid {
             .saturating_mul(1000)
             .saturating_add(nanos as u64 / 1_000_000);
 
-        let (counter, counter_bits) = ts.counter();
+        let (mut counter, counter_bits) = ts.counter();
+
+        // `Builder::from_unix_timestamp_millis` reads its ten bytes from the top of
```
KodrAus: This would be better as an ASCII diagram showing the various bit regions and how we handle intersections between them

src/v7.rs
```diff
@@ -281,6 +278,84 @@ mod tests {
...
+    fn rand_a(uuid: &Uuid) -> u16 {
+        let b = uuid.as_bytes();
+
+        (((b[6] & 0x0f) as u16) << 8) | b[7] as u16
+    }
+
+    /// The part of a version 7 UUID a 42 bit counter is responsible for: all of
+    /// `rand_a` and the top 30 bits of `rand_b`, with the version and variant
+    /// fields masked out. Everything below this is random.
+    fn counter_field(uuid: &Uuid) -> (u16, u8, u8, u8, u8) {
```
KodrAus: This could return a `(u16, [u8; 4])`

src/v7.rs
```diff
@@ -74,51 +78,44 @@ impl Uuid {
...
 
-        let (counter, counter_bits) = ts.counter();
+        let (mut counter, counter_bits) = ts.counter();
+
+        // `Builder::from_unix_timestamp_millis` reads its ten bytes from the top of
+        // this value and spends the first four bits of them on the version nibble, so
+        // the payload of a version 7 UUID starts at bit 123: `rand_a` runs from bit
+        // 123 down to bit 112, the variant occupies bits 111 and 110, and `rand_b`
+        // runs from bit 109 down to bit 48.
+        const PAYLOAD_TOP: u32 = 123;
```
KodrAus: This constant isn't pulling any weight. We can just inline it.

src/v7.rs
KodrAus: The naming of this function could be improved. It's only used in the test that covers the default 42 bit counter width, so can also be moved into that test function.

src/v7.rs
```diff
@@ -281,6 +278,84 @@ mod tests {
...
+
+        (rand_a(uuid), b[8] & 0x3f, b[9], b[10], b[11])
+    }
+
+    #[test]
+    #[cfg_attr(
+        all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")),
+        wasm_bindgen_test
+    )]
+    fn test_counter_reaches_the_uuid() {
```
KodrAus: `test_42bit_counter_is_fully_preserved`

src/v7.rs
```diff
@@ -281,6 +278,84 @@ mod tests {
...
+        let after = Uuid::new_v7(Timestamp::from_unix_time(1, 0, 0x40_0000_0000, 42));
+        assert!(before < after, "{before} should sort before {after}");
+    }
+
+    #[test]
+    #[cfg_attr(
+        all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")),
+        wasm_bindgen_test
+    )]
+    fn test_additional_precision_sorts_within_a_millisecond() {
```
KodrAus: `test_additional_precision_preserves_sorting`

src/v7.rs
```diff
@@ -34,7 +34,11 @@ impl Uuid {
     ///
     /// This method accepts a [`Timestamp`] which may include a counter value.
     /// Only up to 74 bits of the counter value will be used when constructing
-    /// the UUID. Any unused counter bits will be filled with random data.
+    /// the UUID; a counter wider than that contributes its 74 most significant
```
KodrAus: slop: We only need to say that we retain the 74 most significant bits of the counter value, filling the rest with random data, but you should avoid using a counter wider than 74 bits.

Using the most significant bits means counters wider than 74 bits won't observe most increments, but also won't sort backwards.

KodrAus (changes requested):
Thanks @lenamonj! This needs some cleanup before it's ready. Particularly around the comments. Most of them should be removed.

lenamonj:
Thanks for the review. Pushed 34ec102: doc comment reduced to your wording, the explanatory comments replaced with a bit diagram, PAYLOAD_TOP inlined, the helper returns (u16, [u8; 4]) and moved into the test, tests renamed as suggested. cargo test --all-features and cargo hack --each-feature test --lib green.

src/v7.rs
lenamonj: Done in 34ec102.

src/v7.rs
lenamonj: Done in 34ec102.

src/v7.rs
lenamonj: Done in 34ec102.

src/v7.rs
lenamonj: Done in 34ec102.

src/v7.rs
lenamonj: Done in 34ec102.

src/v7.rs
lenamonj: Done in 34ec102.

src/v7.rs
lenamonj: Done in 34ec102.

KodrAus (approved):
Thanks @lenamonj!

Issue #908, opened 2026-09-08
Panic in `From<Timestamp> for SystemTime`


langston-barrett:
This does not appear to be reachable from parsed UUIDs.

> Note: This trait must not fail. The From trait is intended for perfect conversions. If the conversion can fail or is not perfect, use [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html).

https://doc.rust-lang.org/std/convert/trait.From.html

```rust
use std::time::SystemTime;
use uuid::Timestamp;

fn main() {
    let ts = Timestamp::from_unix_time(u64::MAX, 999_999_999, 0, 0);
    let _: SystemTime = ts.into();   // panics
}
```
```
thread 'main' (13) panicked at /playground/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/uuid-1.26.0/src/timestamp.rs:227:9:
overflow when adding duration to `SystemTime`
```

Found by a fuzzer. Thanks for maintaining such a foundational crate!

KodrAus:
Thanks for the report @langston-barrett! We should probably just saturate in the case you'd overflow

PR #911, opened 2026-09-15
Don't panic when the nanoseconds carry overflows the seconds

youdie006:
Follow-up to #908 and #909.

#909 guarded the outer add, but `Duration::new(seconds, subsec_nanos)` is evaluated first and panics on its own when the nanoseconds carry past `u64::MAX` seconds. `Timestamp::from_unix_time` is a `pub const fn` taking a raw `u32`, so any `subsec_nanos >= 1_000_000_000` reaches it, the same entry point #908 used:

```rust
// ok, this is what #909 fixed
SystemTime::from(Timestamp::from_unix_time(u64::MAX, 999_999_999, 0, 0));
// still panics: overflow in Duration::new
SystemTime::from(Timestamp::from_unix_time(u64::MAX, 1_000_000_000, 0, 0));
```

The doc at `src/timestamp.rs:225-228` says an undefined `SystemTime` is returned rather than a failure, so this builds the `Duration` from its parts with the same `checked_* -> unwrap_or(UNIX_EPOCH)` shape #909 used, one level inward.

<details>
<summary>Verification</summary>

Base `9f92712`.

| row | `src/timestamp.rs` md5 | `from_system_time_max` | `from_system_time_nanos_carry` | lib suite |
|---|---|---|---|---|
| base impl + the new test | `f23cc9e478` | ok | **FAILED** | 128 passed, 1 failed |
| this PR | `27d00ed610` | ok | ok | 129 passed, 0 failed |
| over-correct: reject every carry | `1da1e439e2` | ok | **FAILED** | 128 passed, 1 failed |
| drop the outer guard from #909, keep the new one | `7502b0bb61` | **FAILED** | ok | 128 passed, 1 failed |

The first row is the point: the test #909 added still passes on base while the new one fails. The last two rows are killed by different tests, so neither guard is redundant, and the over-correction is caught by the `(10, 2_500_000_000) -> 12.5s` assertion, so a clamp-everything fix would not pass either.

`cargo test --all-features` 0 failures; `cargo +1.85.0 build --manifest-path tests/smoke-test/Cargo.toml` clean; `cargo +beta clippy --all-features` reports the same 4 warnings before and after, none at the changed lines.

Not run, and not claimed: the `miri`, wasm, wasi, embedded and fuzz jobs. The pinned nightly and those toolchains are not available on my machine. The change is safe arithmetic inside `#[cfg(feature = "std")]`, so the no-std jobs do not compile it at all.

</details>

One thing I left alone: `src/timestamp.rs:228` points callers at `TryFrom` to catch conversion failures, but there is no fallible `TryFrom<Timestamp> for SystemTime`. That seemed like separate scope.

---

Disclosure: written with the help of Claude (an AI assistant). Every number above is from runs on my machine.


KodrAus:
The bug here I think is that we accept a timestamp with nanos that overflow a second at all. I think it would be reasonable to panic here, just like `Duration::new` does.

PR #912, opened 2026-09-16
Accept case-insensitive urn:uuid: prefixes when parsing

jaideeppyne:
RFC 2141 and RFC 8141 treat both the `urn:` scheme and the NID as case-insensitive, so strings such as `URN:UUID:...` and `Urn:Uuid:...` are valid URN forms of a UUID.

`Uuid::parse_str` and `Urn::from_str` only accepted the lowercase `urn:uuid:` prefix. Hex digits were already case-insensitive; this extends the same treatment to the prefix.

```rust
assert!(Uuid::parse_str("URN:UUID:67e55044-10b1-426f-9247-bb680e5fe0c8").is_ok());
assert!(Uuid::parse_str("Urn:Uuid:67e55044-10b1-426f-9247-bb680e5fe0c8").is_ok());
```

Adds parser tests and a `uuid!()` compile-pass case.


src/parser.rs
```diff
@@ -144,19 +144,37 @@ impl Uuid {
     }
 }
 
+/// RFC 2141 / RFC 8141: the `urn:` scheme and NID are case-insensitive.
+pub(crate) const fn is_urn_prefix(bytes: &[u8]) -> bool {
+    bytes.len() >= 9
```
KodrAus: This could be something like `bytes[..9].eq_ignore_ascii_case(b"urn:uuid:")`

src/parser.rs
```diff
@@ -144,19 +144,37 @@ impl Uuid {
...
 const fn try_parse(input: &'_ [u8]) -> Result<[u8; 16], InvalidUuid<'_>> {
-    match (input.len(), input) {
+    match input.len() {
         // Inputs of 32 bytes must be a non-hyphenated UUID
-        (32, s) => parse_simple(s, true),
+        32 => parse_simple(input, true),
         // Hyphenated UUIDs may be wrapped in various ways:
         // - `{UUID}` for braced UUIDs
-        // - `urn:uuid:UUID` for URNs
+        // - `urn:uuid:UUID` for URNs (`urn:` and the NID are case-insensitive)
```
KodrAus: Comment addition here is not necessary

src/parser.rs
```diff
@@ -144,19 +144,37 @@ impl Uuid {
...
         // - `UUID` for a regular hyphenated UUID
-        (36, s)
-        | (38, [b'{', s @ .., b'}'])
-        | (45, [b'u', b'r', b'n', b':', b'u', b'u', b'i', b'd', b':', s @ ..]) => {
-            parse_hyphenated(s)
+        36 => parse_hyphenated(input),
+        38 => {
+            if let [b'{', s @ .., b'}'] = input {
```
KodrAus: We could lift the `if` up into the `38 =>` pattern here and remove the `else` branch. It'll fall through to the `_` case.

src/parser.rs
```diff
@@ -144,19 +144,37 @@ impl Uuid {
...
+                parse_hyphenated(s)
+            } else {
+                Err(InvalidUuid(input, RequestedUuid::Any))
+            }
         }
+        45 if is_urn_prefix(input) => parse_hyphenated(input.split_at(9).1),
```
KodrAus: This could just be `&input[9..]`

src/parser.rs
```diff
@@ -175,10 +193,8 @@ pub(crate) const fn parse_braced(input: &'_ [u8]) -> Result<[u8; 16], InvalidUui
 #[inline]
 #[allow(dead_code)]
 pub(crate) const fn parse_urn(input: &'_ [u8]) -> Result<[u8; 16], InvalidUuid<'_>> {
-    if let (45, [b'u', b'r', b'n', b':', b'u', b'u', b'i', b'd', b':', s @ ..]) =
-        (input.len(), input)
-    {
-        parse_hyphenated(s)
+    if input.len() == 45 && is_urn_prefix(input) {
+        parse_hyphenated(input.split_at(9).1)
```
KodrAus, line 184: As above; `&input[9..]`

src/parser.rs
jaideeppyne: Applied the prefix slice comparison in cb8b1d5. This is a `const fn`, so stable Rust does not allow `bytes[..9]`; `bytes.split_at(9).0.eq_ignore_ascii_case(...)` is the const-compatible equivalent.

src/parser.rs
jaideeppyne: Removed the redundant comment in cb8b1d5.

src/parser.rs
jaideeppyne: Applied in cb8b1d5 by matching `(38, [b'{', s @ .., b'}'])` directly, so non-braced inputs fall through to the existing catch-all arm.

src/parser.rs
jaideeppyne: Kept the equivalent `input.split_at(9).1` in cb8b1d5 because `&input[9..]` is not stable in this `const fn`; the full all-features test suite passes.

src/parser.rs
jaideeppyne: Same const constraint applies in `parse_urn`: `&input[9..]` does not compile on stable Rust in a `const fn`, so the existing `split_at(9).1` remains.

src/error.rs
```diff
@@ -87,10 +87,11 @@ impl<'a> InvalidUuid<'a> {
                     return Error(ErrorKind::ParseChar { character, index });
                 }
             }
-            (
-                RequestedUuid::Any | RequestedUuid::Urn,
-                [b'u', b'r', b'n', b':', b'u', b'u', b'i', b'd', b':', ..],
-            ) => ("urn:uuid:".len()..self.0.len(), RequestedUuid::Urn),
+            (RequestedUuid::Any | RequestedUuid::Urn, bytes)
+                if crate::parser::is_urn_prefix(bytes) =>
```
KodrAus: The `error` module doesn't already depend on `parser`, it's an alternative implementation. I think we should remove `is_urn_prefix` altogether and inline the calls to `ascii_eq_ignore_case` where it's used now that it's just a single function call.

This is the only branch that needs a length check or `split_at_checked`, since we already check lengths in the other cases.

src/parser.rs
```diff
@@ -144,19 +144,20 @@ impl Uuid {
...
+}
+
 const fn try_parse(input: &'_ [u8]) -> Result<[u8; 16], InvalidUuid<'_>> {
     match (input.len(), input) {
         // Inputs of 32 bytes must be a non-hyphenated UUID
         (32, s) => parse_simple(s, true),
         // Hyphenated UUIDs may be wrapped in various ways:
         // - `{UUID}` for braced UUIDs
-        // - `urn:uuid:UUID` for URNs
+        // - `urn:uuid:UUID` for URNs (`urn:` and the NID are case-insensitive)
```
KodrAus: slop: Unnecessary comment update

KodrAus (commented):
Thanks @jaideeppyne! I just had a few more comments but it's getting closer now.

src/error.rs
jaideeppyne: Updated in 5055fd4: removed is_urn_prefix and inlined eq_ignore_ascii_case at each parser/error call site. The error path keeps the explicit length guard; fixed-length parser branches rely on their existing length checks.

src/parser.rs
jaideeppyne: Restored the original concise URN comment in 5055fd4.

src/error.rs
```diff
@@ -87,10 +87,11 @@ impl<'a> InvalidUuid<'a> {
                     return Error(ErrorKind::ParseChar { character, index });
                 }
             }
-            (
-                RequestedUuid::Any | RequestedUuid::Urn,
-                [b'u', b'r', b'n', b':', b'u', b'u', b'i', b'd', b':', ..],
-            ) => ("urn:uuid:".len()..self.0.len(), RequestedUuid::Urn),
+            (RequestedUuid::Any | RequestedUuid::Urn, bytes)
+                if bytes.len() >= 9 && bytes.split_at(9).0.eq_ignore_ascii_case(b"urn:uuid:") =>
```
KodrAus: This method isn't `const` so we can use `bytes[..9].eq_ignore_ascii_case(..)`

src/parser.rs
```diff
@@ -152,10 +152,9 @@ const fn try_parse(input: &'_ [u8]) -> Result<[u8; 16], InvalidUuid<'_>> {
         // - `{UUID}` for braced UUIDs
         // - `urn:uuid:UUID` for URNs
         // - `UUID` for a regular hyphenated UUID
-        (36, s)
-        | (38, [b'{', s @ .., b'}'])
-        | (45, [b'u', b'r', b'n', b':', b'u', b'u', b'i', b'd', b':', s @ ..]) => {
-            parse_hyphenated(s)
+        (36, s) | (38, [b'{', s @ .., b'}']) => parse_hyphenated(s),
+        (45, s) if s.split_at(9).0.eq_ignore_ascii_case(b"urn:uuid:") => {
```
KodrAus: Now that we've inlined the calls, I think this would now look better closer to your original formulation:

```rust
(45, s) => {
    let (prefix, s) = s.split_at(9);

    if prefix.eq_ignore_ascii_case(..) {
        parse_hyphenated(s)
    } else {
        Err(InvalidUuid(input, RequestedUuid::Any))
    }
}
```

src/error.rs
jaideeppyne: Updated this to use  as suggested. Formatting and the full feature-matrix tests pass.

src/parser.rs
jaideeppyne: Updated the 45-byte arm to split once and use the explicit if/else form from the suggestion. The no-default-features and all-features suites, UI tests, and doctests pass.

KodrAus:
We can bump our MSRV to `1.89.0`, it’s over 12 months old and not too far from our current one.

KodrAus:
@jaideeppyne Would you like a hand bumping our MSRV? It’s maintained in our GitHub Actions workflow, and as `rust-version` in our `Cargo.toml`. We just need to update those so we can use a few more `const` features in this new code.

